PR #51, opened 2015-01-27
fix to run HeidelTime and GUTime

tankle:
1. add some parameters to run HeidelTime and GUtime
2. add the heidelTime and gutime packages 
3. add build.sh to build the project , runSUTime.sh to run time annotate as example
4. To run HeidelTime, one should modify the config.props under the packages/heideltime


gangeli:
This pull request looks like it's pulling in all of HeidelTime? My instinct is that this is a bad idea, unless there is a compelling reason to do so.

tankle:
Yes. I doesn't delete the doc file of HeidelTime , so it has a lot of files been committed.  I will change ti in the next time. So I closed this .

PR #52, opened 2015-01-27
In HeidelTimeAnnotator add a parameter, and change the way to call GUTime in SUTimeMain

tankle:
1. In the new version of HeidelTime , it need a config.props file
2. The result of GUTime result have some tags should be deleted


manning:
Thanks! sorry that this took ages to do.

PR #79, opened 2015-06-20
made test runnable with other default locale

s-bolz:
Made TimingTest runnable with other default locale. This contribution is in the public domain.


manning:
Thanks! Merged.

PR #84, opened 2015-07-23
Performance enhancement for QNMinimizer

snnn:
I try to use QNMinimizer to train a large scale logistic regression model which has millions of features. And I find it spends much time on constructing param range set, which is unnecessary and could be avoided in most cases.


manning:
Thanks Changming, this looks useful!

PR #90, opened 2015-09-06
Fix a bug in OWLQN: use raw gradient to calculate y vectors

snnn:
The gradient used to construct the y_k vectors, should be raw gradient (the gradient of the unregularized loss). See "Scalable Training of L1-Regularized Log-Linear Models" Section 3.2 rule 4.


manning:
Good catch! Thanks a lot, much appreciated. Merged.

PR #91, opened 2015-09-09
This patch primarily adds coref results to the JSONObjectOutputter.

sdrinf:
Specifically:
* Added a coref element collection map to JSONOutputter.print
* Added coref mentions, using the same schema as TextOutputter
* Added a map serializer to JSONWriter.routeObject

This pull request is hereby declared to be Public Domain.


gangeli:
Merged with commit d99bdd0dce76579e4ecd8c2999e19fa9c4fc79e2, though I did then updated the format a bit in b928218980b5c67a7b98e22a465fa56b721b4652.

Thanks!

PR #97, opened 2015-10-25
Minor bug fix in PropertiesUtils

elepedus:
Incorrect conditional logic was causing prefixes to be removed when keepPrefix was true and kept when it was false.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,9 @@
 *.so
 .DS_Store
 .settings/**
+.gradle
+lib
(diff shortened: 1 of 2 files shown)


AngledLuffa:
Looks legit.  I needed to switch the truth value on one of the uses
elsewhere in corenlp to make it work.  Thanks!

gangeli:
Closed with commit 7ffd9ba389b792175791d502f99d619782819039

Issue #1, opened 2013-09-13
Could the project switch to using log4j for logs?


Asimov4:
I see a lot of logs printed to System.out or System.err.
Would it be possible to use a library like log4j http://logging.apache.org/log4j/2.x/ and use log.error, log.warning, log.info, log.debug instead?
That would make it easier for users of the StanfordCoreNLP to manage which logs should be printed by choosing the log level of the project.

dinoboy197:
I'd be very interested in getting involved here. I think having CoreNLP write to a logging API which can route into any particular logging backend of the library user's choice seems like the right approach - this seems pretty industry standard at this point. It's unfortunate to write / maintain / use programs which maintain a nice separation between logging API and logging backend for the purpose of maintaining consistent logging format, level, and log destination, only to include libraries which don't take advantage of these nice logging APIs and just dump unformatted text onto stdout / stderr (I don't mean to pick on CoreNLP here alone - there are hundreds, probably thousands of libraries in Maven Central which blindly dump text onto stdout / stderr).

As for a logging API for CoreNLP to interface with, my recommendation would be to use SLF4J. It doesn't bias the client code (which makes calls to CoreNLP) to use any particular logging backend, nor does it require client code to provide additional backend shims to re-route logging events between backends. For example, by using SLF4j as the logging API, a program including the CoreNLP library needs only to include a single logging backend of their choice, for example logback (the successor to log4j), log4j, commons-logging, or java logging (jul). No logging bridges or shims are required for the end client to use any logging backend that they choose. (Library -> SLF4j API -> SLF4j implementation -> logging backend -> <log destination>)

If CoreNLP used commons-logging as the logging API, then programs which want to use logback, log4j, or jul as the backend would need to include some sort of log event redirection shim (as reckart mentioned above) (typically in the form of SLF4j plus some additional commons-logging to destination backend bridge) which increases logging backend complexity, adds additional library dependencies to the client program, and increases logging event overhead. (library -> commons-logging -> slf4j-commons-logging-shim -> slf4j API -> slf4j implementation -> logging backend -> <log destination>).

[...]

PR #99, opened 2015-10-31
Convert a small number of stdout / stderr calls to slf4j

dinoboy197:
[Based on the decision to switch over to slf4j](https://github.com/stanfordnlp/CoreNLP/issues/1#issuecomment-143598156) per @manning , I thought I would put together a small PR to test the slf4j waters with reviewers.

This uses the Logging class (created previously by another contributor) to retrieve an instance of a Logger. If that isn't desired (you'd rather see just ```Logger logger = LoggerFactory.getLogger()```, please let me know and I'll adjust.

### DIFF
--- a/src/edu/stanford/nlp/io/IOUtils.java
+++ b/src/edu/stanford/nlp/io/IOUtils.java
@@ -20,6 +20,9 @@ import java.util.regex.Pattern;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * Helper Class for various I/O related things.
  *
@@ -34,6 +37,8 @@ public class IOUtils {
 
   public static final String eolChar = System.getProperty("line.separator");
   public static final String defaultEncoding = "utf-8";
+
+  private static Logger logger = LoggerFactory.getLogger(IOUtils.class);
 
   // A class of static methods
   private IOUtils() { }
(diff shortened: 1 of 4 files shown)


J38:
Hi, first of all thanks for the contribution!  We are actually planning on not using that Logger class and just following the standard slf4j pattern ; the pattern we're going to follow is is edu.stanford.nlp.classify.SVMLightClassifierFactory as an example.  Slowly but surely over time we will be converting everything over to this new logging method.

dinoboy197:
> We are actually planning on not using that Logger class and just following the standard slf4j pattern referenced here: http://www.slf4j.org/manual.html ; the pattern we're going to follow is in edu.stanford.nlp.classify.SVMLightClassifierFactory as an example.

Perfect! I actually prefer that style too, but i just wasn't sure since this Logging class existed...

Let me adjust my PR.

dinoboy197:
@J38  Ping! Just checking back in on this PR.

J38:
Hi I'll merge this today, it looks great!  Thanks again for the help!  If you are interested in helping more we'd greatly appreciate it!  Our plan is to hold off on any more logging changes until we release 3.5.3 and finalize the logging plans.  But if you'd like to convert statements, you could convert a package at a time, and then I'll review what you did and merge it in!  Once again thanks for your help and interest in Stanford CoreNLP!

dinoboy197:
> But if you'd like to convert statements, you could convert a package at a time, and then I'll review what you did and merge it in!

You bet - do you mean just converting ```System.out.println``` and ```System.err.println``` into corresponding slf4j statements? I'm happy to keep working on that :)

J38:
Yes I think our first goal should be to change the System.out.println and System.err.println statements, and then eventually change everything else.  Thanks again!

manning:
Thanks. Merged.

PR #100, opened 2015-11-03
Change to Diamond <> operators; replace for loops with foreach

lababidi:
Mainly a language cleanup.


manning:
Thanks a lot!  (Except that you seem to have reformatted MaxentTagger, which I undid....)

PR #102, opened 2015-11-14
System logger to Slf4J conversion for wordseg package

dinoboy197:
Per @J38 in https://github.com/stanfordnlp/CoreNLP/pull/99#issuecomment-154529499, I'm starting to convert packges one by one from ```System.out.println``` and ```System.err.println``` into SLF4J statements.

### DIFF
--- a/src/edu/stanford/nlp/wordseg/ChineseDictionary.java
+++ b/src/edu/stanford/nlp/wordseg/ChineseDictionary.java
@@ -4,6 +4,9 @@ import java.io.*;
 import java.util.*;
 import java.util.regex.Pattern;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import edu.stanford.nlp.io.IOUtils;
 import edu.stanford.nlp.io.EncodingPrintWriter;
 import edu.stanford.nlp.io.RuntimeIOException;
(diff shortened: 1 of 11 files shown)


J38:
Thanks for your help on this!  If you make changes to other packages I am happy to review them and then merge them in.  I think doing this on a package by package basis makes the most sense.  Thanks again!

J38:
FYI we've released 3.6.0 which includes slf4j-api.jar and slf4j-simple.jar in the distribution.  In the pom.xml though we are omitting slf4j-simple.jar from the dependencies so people can choose whatever logging they want to use for slf4j.

dinoboy197:
> I think doing this on a package by package basis makes the most sense. And you can see some of your code changes in release 3.6.0! Thanks again for the help!

You're welcome! I'll continue converting on a package by package basis when I have time.

> In the pom.xml though we are omitting slf4j-simple.jar from the dependencies so people can choose whatever logging they want to use for slf4j.

Perfect.

craigryan:
So grateful you're adding logging. I'm trying out 3.6.0 and the sentiment examples using annotators "tokenize, ssplit, parse, sentiment" and the only output I'm still seeing is "done [10 sec].". Looks to be coming from edu.stanford.nlp.util.Timing:324

public void done() {
    System.err.println("done [" + toSecondsString() + " sec].");
  }

I'd really appreciate if you could add the 'utils" package to your list of updates as I presume these classes are consumed by other packages that already have logging in place.

thanks!

BTW, if it's any use to anyone either of the following should turn off nlp logging if using the default logger:

        System.setProperty("org.slf4j.simpleLogger.log.edu.stanford.nlp", "error");
    	System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", "error");

dinoboy197:
> I'd really appreciate if you could add the 'utils" package to your list of updates as I presume these classes are consumed by other packages that already have logging in place.

I'd be happy to help out where I can. I'd be really interested in your exact use case - perhaps take a look at how you use CoreNLP in your code to see what makes sense.

craigryan:
Yep sure Taylor, my opensource project (sourceforge: crgrep) is a search
tool and makes use of various third party libraries depending on the type
of resource being searched. I'm working on adding a new sentiment (mood)
option in order to take text extracted from a resource and filter the
search results by the expected sentiment. So 'crgrep --mood negative
'product' surveys/*' will search all survey files that contain references
to 'product' and express a negative sentiment.

Having spurious output from any third party library appearing in the
results is confusing without any kind of context so I need the search tool
to control the way in which results and errors are reporting in a way that
makes sense.

Craig.

PR #114, opened 2015-12-16
Fixes concerning MUC6 and MUC7 coreference resolution input files

peschue:
These changes make it possible to use dcoref with MUC6 and MUC7 files for coreference resolution as distributed by LDC. The existing code had problems with both files.

(I declare that this contribution is in the public domain.)


peschue:
I found out that the fixes circumvent crashes but some results are unexpected. I will investigate and open another pull request if I am confident that everything is really fixed wrt MUC6/MUC7.

PR #115, opened 2015-12-17
Fixed -unkWord option in RNNOptions.java

mschnoell:
The -unkWord option is usable now ;-)


AngledLuffa:
Oops, how embarrassing.  Thanks for reporting that!

We really should have an arg parsing library (ideally one which can handle
enums)

gangeli:
Technically, we do have an arg parsing library (http://nlp.stanford.edu/nlp/javadoc/javanlp/), and it even handles Enums. I'm just bad at advertising it...

AngledLuffa:
What's it called?  It doesn't show up on that page, at least not if you
search for "arg"

gangeli:
Ack, I forgot Javadoc has iframes: http://nlp.stanford.edu/nlp/javadoc/javanlp-3.6.0/edu/stanford/nlp/util/Execution.html. The OpenIE code uses it, for example: https://github.com/stanfordnlp/CoreNLP/blob/master/src/edu/stanford/nlp/naturalli/OpenIE.java#L99

PR #118, opened 2015-12-28
Chinese coreference resolution API example

lizekui:
Because the API code in http://stanfordnlp.github.io/CoreNLP/coref.html
and http://nlp.stanford.edu/software/dcoref.shtml are not runable under
Chinese text instances test, I modified the properties file, which is
extracted from stanford-chinese-corenlp-2015-12-08-models.jar file.
Finally our code is successfully run under Chinese.

============================
1) ADD properties file under edu/stanford/nlp/hcoref/properties/zh-coref-default.properties

2) ADD java file under itest/src/edu/stanford/nlp/hcoref/HcorefChineseAPIExampleTest.java


J38:
Thank you for your help!

PR #144, opened 2016-02-15
nndep Classifier and Config: Prevent fine-tuning with flag.

cinjon:
Add a flag (default: true) for performing the word embedding gradient update. The main purpose is for those who want to provide their own embeddings and not fine-tune them.


manning:
Looks good, thanks!

manning:
Merged.

cinjon:
This implementation has a problem actually in that it deals poorly with any word embeddings in the training data that aren't in the provided embedFile. They are initialized to random vectors and then are neither treated as -UNKNOWN- nor updated in the gradient step.

I think that the more faithful solution would be to treat them as -UNKNOWN-, but that's not the intention of the code. An alternative would be to selectively update those embeddings only.

What do you think? I can make the change.

manning:
Maybe @hans or @danqi could give a more expert opinion, but it would seem okay to me to treat them as UNK.

PR #166, opened 2016-03-30
Updated charniak parser wrapper to be able to work with latest

reshet:
bllip-parser executable parse-50best.sh
Details: https://github.com/BLLIP/bllip-parser


J38:
Ok I have added this thanks for the update!

PR #188, opened 2016-05-12
Add stackedNer key to AnnotationLookup to allow hybrid NER training

drndos:
This additional key allows for adding a stackedNer to a properties map for CRFClassifier:
`properties.put("map", "word=0,stackedNer=1,tag=2,lemma=3,answer=4");`
So stackedNer feature in CRFClassifier can be fully used for training statistical model.

I hereby declare this one line of code public domain or whatever license you need.


manning:
Thanks!

PR #190, opened 2016-05-16
gradle wrapper added.

alishir:


manning:
Thanks, but will update to more recent gradle.

PR #191, opened 2016-05-17
Added support for InputStream constructor in MaxentTagger

TobiasNorlund:
I have added constructors for inputting the trained model as an `InputStream` in `MaxentTagger`. Currently it only allows a model file path string. 

It is useable it the model files are stored somewhere else (e.g the cloud or in a db) and not reachable by a file path string.

The edits in this PR are intended for the public domain.


AngledLuffa:
Seems reasonable.  Thanks for sharing.

PR #204, opened 2016-06-10
Added functionality for contains method in SUTime.Range and JUnit-Test

Erechtheus:
Implemented the "contains" method for SUTime.Range, which checks if another range is within this range. Previously the method always returned false. Also added JUnit test...


manning:
Thanks!

PR #209, opened 2016-06-18
Switches default truecase model to model from default distribution

rationalism:
The model "truecasing.fast.caseless.qn.ser.gz" is set as the default for the truecase annotator, but this model isn't included in the default CoreNLP distribution (http://stackoverflow.com/questions/35574874/core-nlp-truecaseannotator-not-found). This changes the default model to "truecasing.fast.qn.ser.gz", which is included.


manning:
I think we made a mistake in terms of which model was included in the jars.

We should have kept `truecasing.fast.caseless.qn.ser.gz` in the jar (and it is still specified in the code as the default truecasing model).

Really,  `truecasing.fast.qn.ser.gz` is not an arbitrary truecaser, but just an upcaser....

You can currently find the caseless model in the "english" models jar. For the 3.7 release, we will change what is in the default jar.

So, thanks for pointing out the problem, but therefore we won't merge this.

PR #211, opened 2016-06-24
batch endTree index bug fix

jhujhuang:
I think the original code meant to have `(startTree + batchSize > shuffledSentences.size())` as the if condition, but instead had `(endTree + ...)`. This may make the second to last batch to have a wrong `endTree` index, thus including the data from the last batch, and these data would be used twice.


AngledLuffa:
Looks correct to me.  I can't find your email address, but if you send it to me, I can make sure you get properly credited.  (Or not, if you prefer.)

AngledLuffa:
Merged

PR #222, opened 2016-07-15
Use correct CoNLL-X POS field

nhynes:
According to the [CoNLL-X spec](http://ilk.uvt.nl/conll/), the POS is in the fourth field whereas the coarse-grained POS is in the third, which what is currently extracted.

This PR is also consistent with [parser.nndep.Util.loadConllFile](https://github.com/stanfordnlp/CoreNLP/blob/master/src/edu/stanford/nlp/parser/nndep/Util.java#L145) which defaults to using the fine-grained POS.

Since the POS is equivalent to the CPOS when unavailable, this change should provide more utility to downstream applications without introducing many backward incompatibilities.


manning:
Seems valid! Thanks!

PR #225, opened 2016-07-18
bugfix: NERServer not closing connection on error

antjkennedy:
NERServer would not close the connection to the client should the
data enrichment process fail. Added finally block to close connections
to the client.

Now catching OOM errors as it is easy to
misconfigure the server when you are unsure how it will be used by
consumers.

Finally DEBUG messages can be turned on by setting the 'NERSERVER_DEBUG'
environment variable to "true". This allows debug messages to be
turned on without recompiling the server.

commited as 'de minimis' as there are no significant changes here, just
the addition of a finally block and some logic for getting the
environment variable.


AngledLuffa:
Will be happy to merge this, thanks.  If you want to add your email address to your account you can have proper credit.

antjkennedy:
Excellent! My email should now be visible on my profile.

AngledLuffa:
Great!  Merged.

PR #250, opened 2016-08-21
Remove Google Analytics code from local Server demo installations.

willmoy:
CoreNLP Server demo includes Google Analytics code.  It uses the Google Analytics ID of http://corenlp.run/. This is presumably an error.

This patch removes GA code from the template.


manning:
Thanks, yes, it really shouldn't be there in the github version.

PR #253, opened 2016-08-28
Fix NPE for ColumnDataClassifier initialization from model (enableassertions case)

UshakovVasilii:
`flags` here always null. Problem is reproduced with `-enableassertions` java oprion


AngledLuffa:
Merged

AngledLuffa:
Thank you!

PR #254, opened 2016-09-01
Tregex Server Support

sbryfcz:
I have added support for running tregex queries on a CoreNLP Server. 

One can make a POST request similar to semgrex:
```
http://localhost:9000/tregex?pattern=NP<NN=noun
```
with body:
```
The quick brown fox jumps over the lazy dog.
```

and expect the result to be in the form:
```
{
  "sentences": [
    {
      "0": {
        "match": "(NP (DT The) (JJ quick) (JJ brown) (NN fox))\n",
        "namedNodes": [
          {
            "noun": "(NN fox)\n"
          }
        ]
      },
      "1": {
        "match": "(NP (DT the) (JJ lazy) (NN dog))\n",
        "namedNodes": [
          {
            "noun": "(NN dog)\n"
          }
        ]
      }
    }
  ]
}
```

I tried to model the code after what already exists in the server application but am open to suggestions and critique. I'm just getting familiar with CoreNLP.

While I think this will help people experiment with Tregex, I plan on continuing to work on this and create a text box to play with these expressions just like you can with Semgrex and Tokensregex in the client application.

### DIFF
--- a/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
+++ b/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
@@ -12,6 +12,13 @@ import edu.stanford.nlp.ling.tokensregex.TokenSequencePattern;
 import edu.stanford.nlp.semgraph.SemanticGraphCoreAnnotations;
 import edu.stanford.nlp.semgraph.semgrex.SemgrexMatcher;
 import edu.stanford.nlp.semgraph.semgrex.SemgrexPattern;
+import edu.stanford.nlp.trees.Tree;
+import edu.stanford.nlp.trees.TreeReader;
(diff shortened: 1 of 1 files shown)


manning:
Thanks! Nice to have! But I think I might move the TregexHandler to a separate class, since it seems a bit of a mistake to keep stuffing more and more stuff into StanfordCoreNLPServer.

manning:
It would indeed be lovely if there were also a textbox for playing with tregex in the web interface.

sbryfcz:
Yeah. The StanfordCoreNLPServer class is quite large. I didn't want to break it up in case the team wanted it that way. Just trying to follow suit.

And I completely agree about the text box. I started working on it, but haven't had time to really complete it. Happy to give it a shot if you'd like.

PR #257, opened 2016-09-04
batch endTree index fix

jhujhuang:
parser.dvparser.DVParser has the same endTree index bug that sentiment.SentimentTraining had #211

### DIFF
--- a/src/edu/stanford/nlp/neural/rnn/RNNCoreAnnotations.java
+++ b/src/edu/stanford/nlp/neural/rnn/RNNCoreAnnotations.java
@@ -7,14 +7,20 @@ import edu.stanford.nlp.ling.CoreLabel;
 import edu.stanford.nlp.ling.Label;
 import edu.stanford.nlp.trees.Tree;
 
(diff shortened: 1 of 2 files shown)


src/edu/stanford/nlp/parser/dvparser/DVParser.java
```diff
@@ -154,7 +154,7 @@ public void train(List<Tree> sentences, IdentityHashMap<Tree, byte[]> compressed
         // the list
         int startTree = batch * op.trainOptions.batchSize;
         int endTree = (batch + 1) * op.trainOptions.batchSize;
-        if (endTree + op.trainOptions.batchSize > shuffledSentences.size()) {
+        if (endTree > shuffledSentences.size()) {
```
AngledLuffa: I *think* that the subsequent call to `shuffledSentences.subList(startTree, endTree)` will keep the last sentences from being repeated.  However, it certainly is ugly written the way it is, so I will merge this change.

AngledLuffa:
Merged.

PR #262, opened 2016-09-12
Correct check for undefined variable

mathiasrw:
The code was comparing the result of `typeof` with a variable named `undefined`.

As typeof returns a string it should compare to `'undefined'`.


manning:
Thanks! I'm no javascript expert, but it seems like most of the online stuff seems to also recommend using strict equality (===), so maybe I'll also change it to that....

mathiasrw:
Good point!

PR #277, opened 2016-10-14
add missing WEEKDAY variable

garfieldnate:
SUTime used the variable "WEEKDAY" in some expressions, but it never 
defined that expression. This resulted in the somewhat frequent warning
Unknown variable: WEEKDAY. Note that the Timex2 standard does not 
have a weekday symbol (though it has a weekend symbol), and my choice
of "WD" is not standard (though it seems like an obvious choice).


J38:
Ok this has been added in.

PR #279, opened 2016-10-19
Server: customize NER model path

thammegowda:
I wasn't able to set custom ner.model path to the Server, so I modified few lines of code to get it working.

we can now set `-Dner.model` system property at runtime.

```
java -cp .:* \ 
 -Dner.model="ner-model.ser.gz,edu/stanford/nlp/models/ner/english.all.3class.distsim.crf.ser.gz"  \ 
edu.stanford.nlp.pipeline.StanfordCoreNLPServer
```

This is a minor change involving 6 additions.

### DIFF
--- a/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
+++ b/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
@@ -69,6 +69,7 @@ public class StanfordCoreNLPServer implements Runnable {
   protected static String defaultAnnotators = "tokenize,ssplit,pos,lemma,ner,parse,depparse,mention,coref,natlog,openie,regexner,kbp";
   @ArgumentParser.Option(name="preload", gloss="Cache the following annotators on startup")
   protected static String preloadedAnnotators = "";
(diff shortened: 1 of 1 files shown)


J38:
I am going to work on how default properties are given to the server when it is started, and I will probably just add code so that the server can take in a properties file and command line properties the way StanfordCoreNLP's main() method does.

For the time being, if you want to specify the "ner.model" that is used you can do this with the request.

Here is an example ``` wget --post-data 'The quick brown fox jumped over the lazy dog.' 'localhost:9000/?properties={"annotators":"tokenize,ssplit,pos,lemma,ner","outputFormat":"json", "ner.model":"edu/stanford/nlp/models/ner/english.all.3class.distsim.crf.ser.gz"}' -O -```

thammegowda:
@J38 Thanks for your fix in v3.7.0 👍 
IMO, the workaround used to expose server-side paths to client programs, and hence it was not OK!!

I confirm, your fix works:
```
java -mx5g -cp "*" edu.stanford.nlp.pipeline.StanfordCoreNLPServer -port 9000 \
 -serverProperties StanfordCoreNLP.properties
```

PR #283, opened 2016-10-24
Added case-sensitive parameter for TokensRegrex calls to CoreNLP Server

lteacy:
Added support for case-sensitive parameter in TokensRegex calls to StanfordCoreNLPServer requests.

If case-sensitive = false, then TokensRegrex matches will be case insensitive.
If case-sensitive parameter is not specified or set to any other value, then the default case sensitive behaviour is maintained.

Happy to make this contribution in the public domain.


J38:
So the TokensRegex annotator already has a way to be case-sensitive or not.

Here is an example request (note the ignorecase=true):

wget --post-data 'john kerry was born in colorado.' 'localhost:9000/?properties={"annotators":"tokenize,ssplit,pos,lemma,ner,regexner","regexner.mapping":"ignorecase=true,validpospattern=^(NN|JJ).*,edu/stanford/nlp/models/kbp/regexner_caseless.tab;edu/stanford/nlp/models/kbp/regexner_cased.tab", "outputFormat":"json"}' -O -

PR #284, opened 2016-10-24
Semgrex Server requests now always run against Enchanced++ dependencies.

lteacy:
Previously, looks like Semgrex Server requests were only run against Enhanced++ dependencies if the filter parameter was set to True. Otherwise, looks like the older Enhanced graph is still being used?

Changed this so that Enhanced++ dependencies are used in all cases.


J38:
Thanks for finding this issue!  The pull request has been merged!

PR #289, opened 2016-10-29
Burton new pom for building

burtonator:
Took pom-full.xml, moved it to the root, then updated the dependencies so we can build with maven.  

Works well so far... now I can try to resolve some issues we've been facing internally.  

this supports building with maven without the .jars having to be in lib.


manning:
Thanks!  Merging with a little updating for 3.7.0.

burtonator:
Awesome.  Thanks!

PR #295, opened 2016-11-01
Add info about using server in a shell script

lenny-t:
Add a note and example for using the CoreNLP server inside a shell script.


manning:
Thanks! Merging.

PR #325, opened 2016-12-22
Add logging to SentimentCostAndGradient.

sarahsharp:
When adding hundreds of new sentences to a sentiment training file, it
is highly annoying to not know which sentence has a mistake in it. The
PennTreeReader class provides some nicer logs that print the Penn tree
when it detects sentences with mismatched parenthesis, so I added
similar logging to SentimentCostAndGradient. There's no need to check
whether tree is null, since the if statements already dereference the
variable.

I certify that this is a bug fix of 6 lines, and thus not requiring me
to sign the Stanford CoreNLP CLA.

Signed-off-by: Sarah Sharp <sharp@otter.technology>


manning:
Thanks!

Issue #263, opened 2016-09-13
ChineseSegmenterAnnotator / ChineseStringUtils NullPointerException in multi-threaded use


pauldoo:
When using `ChineseSegmenterAnnotator` in a multi-threaded environment a NullPointerException can sometimes occur.  The root cause seems to be that `ChineseStringUtils.processColons()` is not thread safe due to lazy initialisation of static members.

Current master revision can throw an NPE on line 300:
https://github.com/stanfordnlp/CoreNLP/blob/6ea6a2e9e9b58b41827ebbf581ca2239aeebb52f/src/edu/stanford/nlp/wordseg/ChineseStringUtils.java#L300

The issue is the (thread-)unsafe initialisation of `colonsPat` a few lines earlier.  The same issue appears to be present for `colonsWhitePat`.

manning:
Closing assuming fixed!

PR #337, opened 2017-01-10
Refactored ChineseStringUtils.java to remove multithreading issues

KellenSunderland:
Hello, I've done a refactor of the ChineseStringUtils.java to address some multithreading issues that I've run into.  My problems were pretty similar to the ones reported in #263 .  I tried to refactor the code to maintain the lazy-loading of regex patterns, but also to make everything threadsafe.   

One note of caution: I'm not an expert in Chinese Segmentation, and there didn't seem to be many unit tests covering this code, so it would be great if someone could verify that there's no behavior change here.

### DIFF
--- a/src/edu/stanford/nlp/wordseg/ChineseStringUtils.java
+++ b/src/edu/stanford/nlp/wordseg/ChineseStringUtils.java
@@ -1,28 +1,32 @@
 package edu.stanford.nlp.wordseg;
 
-import java.io.File;
+import edu.stanford.nlp.io.EncodingPrintWriter;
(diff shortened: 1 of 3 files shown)


KellenSunderland:
All new/modified code in this PR is in the public domain.

manning:
Thanks a lot!

manning:
p.s. It's far from clear that the lazy loading of regex in this class is really in any way useful rather than just a needless complication....

KellenSunderland:
Thanks for the merge!  I totally agree that the lazy initialization is likely unneeded complication.  I can try and clean it up using simple static initialization if no-one objects to losing lazy loading.

PR #356, opened 2017-02-13
Fix jarfile dependency issue for Makefile compilation

KennyWibowo:
Jar dependencies aren't properly included in the corenlp target, causing a bunch of missing imports compiler errors.


KennyWibowo:
This closes #357.

J38:
Our desire with this Makefile is to be able to rebuild the jar containing the .class files, so it is intentional to not include the dependencies in the stanford-corenlp-3.7.0.jar

PR #385, opened 2017-03-18
Update Neural CorefSystem to retain singletons

henryYHC:
After the change, the *.coref.predicted.txt will retain singleton mentions. Or else, they will be filtered out regardless of the variable "removeSingletonClusters".


J38:
So the point of this method is to write CoNLL output for running evaluations on the CoNLL 2012 coreference data, which would mean you always want to get rid of singleton mentions.  I think the intent here was to run CoNLL evaluations and hard code removing singletons.

If you just want to run coreference on general text with the `coref` annotator and get singleton mentions the removeSingletonMentions setting should work.

PR #386, opened 2017-03-18
Hi, I fixed https://github.com/Felix-Pobaby/CoreNLP/issues/1

Felix-Pobaby:
Please refer the issue [TokensRegexHandler not properly uses pipeline to annotate](https://github.com/Felix-Pobaby/CoreNLP/issues/1)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 *.so
 .DS_Store
 .settings/**
+.classpath
+.project
+target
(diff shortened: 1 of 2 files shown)


Felix-Pobaby:
change the pull request source branch.

gangeli:
Thanks for the fix! My concern here is if there are two users of the server with different pipelines, one of them can clobber the properties of the other in between two calls. For instance, this is the server running at corenlp.run, and multiple users are making requests against it at a time.

Really, at some point the tokensregex and semgrex endpoints should just take as input annotated documents (e.g., as proto or json) and run the just the patterns on them; or, there should be an option to run tokensregex/semgrex in conjunction with the main annotation pipeline. But both of these are a bit of work, and currently low priority, at least for me.

Felix-Pobaby:
You are right. When multi-users access the corenlp hanlder, then their pipelines will interfere with each other if without controll.
So, We should find a way to cache the pipeline in the sesssion.

Something difficult: CoreNLP's pipeline is too big to store in the memory if used by many user. As the server mode the CoreNLP uses is not a good framework. My group can redesign and develop a good enough web service framework for CoreNLP.

Right now, we are just using the limited cached pipeline to prcatice TokensRegex, so I just merge the bug fixing in my own fork.

Thanks for your apply!!

PR #388, opened 2017-03-18
Find bugs and fixed: plus is not properly replace as wrong regex

Felix-Pobaby:
Please refer ["wrong process back slash in TokensRegex "](https://github.com/Felix-Pobaby/CoreNLP/issues/2)
Thanks very much!

### DIFF
--- a/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
+++ b/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
@@ -212,7 +212,7 @@ public class StanfordCoreNLPServer implements Runnable {
       String query = uri.getQuery();
       String[] queryFields = query
           .replaceAll("\\\\&", "___AMP___")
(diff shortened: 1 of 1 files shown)


J38:
Thanks for finding this issue!

J38:
Hi I wanted to get some feedback from @gangeli and @Felix-Pobaby 

So I changed the regex as requested.  But I'm not finding the url @Felix-Pobaby suggests to be the correct one to use.  What I see working properly is escaping "+"'s before encoding.

Let's take the use case of pattern = /\S+/ , and I'll show how I make the proper url with Python's urllib library.

1.) Escape all +'s in the desired pattern: `"/\S\+/"`
2.) `urllib.urlencode({"pattern": "/\S\+/"})` creates `pattern=%28%2F%5CS%5C%2B%2F%29`
3.) issue this request:
`   wget --post-data '...' http://localhost:9000/tokensregex?pattern=%28%2F%5CS%5C%2B%2F%29`

PR #389, opened 2017-03-18
target dir and .classpath .project should be ignored in git

Felix-Pobaby:
Please refer [target dir and .classpath .project should be ignored.](https://github.com/Felix-Pobaby/CoreNLP/issues/3)
Thanks very much!


J38:
We want .project and .classpath backed up in git.  I will add target to .gitignore though.

PR #392, opened 2017-03-22
Make JSONWriter constructor public

uiri:
JSONWriter is a public class but its constructor is currently
private. This renders it completely unuseable for any practical
purpose outside of JSONOutputter and JSONOutputter's test.

This makes its constructor public. This allows it to be used in cases
where e.g. the JSONOutputter is subclassed.


manning:
Thanks for the contribution, merging it. Though if people start using JSONWriter much, it'd probably make sense to promote it to being a top-level class rather than an inner class.

PR #423, opened 2017-04-28
Support passing parameters on page launch, via window.location.hash

julien-c:
The need for this was briefly discussed on [Stack Overflow](http://stackoverflow.com/questions/43663335/use-get-params-to-provide-the-web-interface-with-a-specific-text-to-annotate).

I would like to link to my instance of CoreNLP server, with a specified text and specified set of annotators (i.e. without having to paste the text then click on Submit).

Rather than using GET params it's easier (as the whole frontend is static) to just parse parameters in JS.

Example URL: 
```
http://corenlp.run/#text=foo%20bar&annotators=pos,lemma,ner
```

### DIFF
--- a/src/edu/stanford/nlp/pipeline/demo/corenlp-brat.js
+++ b/src/edu/stanford/nlp/pipeline/demo/corenlp-brat.js
@@ -1066,6 +1066,41 @@ $(document).ready(function() {
     return false;
   });
 
+  // Support passing parameters on page launch, via window.location.hash parameters.
+  // Example: http://localhost:9000/#text=foo%20bar&annotators=pos,lemma,ner
+  (function() {
+    var rawParams = window.location.hash.slice(1).split("&");
+    var params = {};
+    rawParams.forEach(function(paramKV) {
+      paramKV = paramKV.split("=");
+      if (paramKV.length === 2) {
+        var key   = paramKV[0];
+        var value = paramKV[1];
+        params[key] = value;
+      }
+    });
+    if (params.text) {
+      var text = decodeURIComponent(params.text);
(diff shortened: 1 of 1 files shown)


julien-c:
Thanks @gangeli – when do you guys usually rebuild the production jar? (I'm building my own jar from the tip of `master` and using it in production but it seems to be less stable – meaning, locks up the jvm more frequently... – than usual)

More generally, any tips on deployment of production-grade corenlp instances?

gangeli:
It shouldn't be less stable off of HEAD. Releases usually just pick a point that looks good enough to release, freeze it + test it a bit more thoroughly, and then put it out into the world.

What do you mean by locking up the JVM? We did tweak the annotator pool implementation a bit ago, which could be causing problems.

RE production-grade CoreNLP: at Eloquent (my current company) we've had success just putting the server in a Docker onto a Kubernetes cluster. UptimeRobot pegs us at 99.85% uptime, with a 100% success rate in the last month on an [hourly] integration test.

gangeli:
Can you try again? There were indeed some bugs with the annotator pool, and it seems healthier now. Works on corenlp.run at least: http://corenlp.run/#text=foo%20bar&annotators=pos,lemma,ner

J38:
@gabor @julien-c I just wanted to double check, does everyone think this code change will work with the demo ?  If so I will merge this pull request.

julien-c:
I think this was already merged. (not the pull request itself but its content)

J38:
Oh, does it work for you when you just use our latest code ?

J38:
Oh I guess it does, cool!

PR #430, opened 2017-05-13
Bump version number for EJML.

vincentk:
In particular, the maven group id and the package prefixes are now the
same, thereby eliminating potential runtime errors for conflicting
dependencies.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,9 +97,9 @@
     </dependency>
 
     <dependency>
-      <groupId>com.googlecode.efficient-java-matrix-library</groupId>
-      <artifactId>ejml</artifactId>
-      <version>0.23</version>
+      <groupId>org.ejml</groupId>
+      <artifactId>all</artifactId>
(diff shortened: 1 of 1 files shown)


J38:
If we upgrade EJML, we also would want to upgrade the jar we distribute in the lib folder.  I will talk with Chris to see if there is any particular reason we are using the older EJML (licensing, etc...?) and why we are using the Google version.  But I think fairly soon I am open to trying to move to a newer version.

manning:
This would be good to do, and I just attempted to. Because of API changes, there are quite a few code changes that have to be made; it's not just bumping the version number.

We will do this eventually, but unfortunately, these code changes mean that a number of models would need to be rebuilt if we upgrade the version. So it's not going to happen for v3.9.

AngledLuffa:
Whenever the next release of CoreNLP occurs, we will include an updated EJML.  This pull request is no longer needed, but thank you.

PR #458, opened 2017-06-08
Use JaFaMa for faster math, and optimize critical code paths

kno10:
These changes substantially cut down the processing time; by several hours when I process all of Wikipedia.
Feel free to benchmark on your own data.

The first commit uses JaFaMa instead of java.lang.Math, which is 2-3x faster for exp, log: http://blog.element84.com/improving-java-math-perf-with-jafama.html
In some places I switched back to log1p, because the runtime of log and log1p in JaFaMa are similar, and log1p offers better precision for small values of x than log(1+x).

The other patches optimize the crucial code around the Viterbi algorithm:

- HotSpot optimizes better if large functions with multiple loops are split into multiple methods (as they can be recompiled independently).
- It pays off to save repeated nested array lookups (e.g. `array[i][j]` in a loop over `j`; move `array_i = array[i]` outside of the loop and use `array_i[j]` inside).
- I also add a cache to avoid recomputing the open tags set in Ttags.

All of these may appear to be trivial changes, but once you benchmark you will see how much this improves the run time.

Processing the first 20000 articles with `tokenize,ssplit,pos`, doing some further processing such as my own lemmatization based on hunspell, and then loading them into a lucene index with the CoreNLP master branch took 08:51 minutes, and with my patches only 04:38 minutes (sloopy benchmark only). I consider this a substantial speedup, because Wikipedia is 5.3 million articles, and it still needed 19 hours to build the full text index, but it used to take almost two days...


J38:
Thanks for contributing this!  I am currently working on finalizing the Stanford CoreNLP 3.8.0 release but when I finish that I will look this over!

J38:
So this looks really cool.  I'm going to try to talk to Chris and see if there are any reasons we shouldn't convert to this, he will have better insight than me!  But the speed ups sound awesome!

kno10:
I rebased my branch onto 0c07f3d0e1d7c0a9b2188a7ba3854245e4b2e7fa because the current git version performs much worse for me (c.f., #584)

The speedup in training I get is 895.0 to 713.4, I.e. about 20% faster training (+ less memory needed).

With these patches:
```
CRFClassifier training ... done [713.4 sec].
...
CRFClassifier tagged 51645 words in 3068 documents at 26189,15 words per second.
         Entity    P       R       F1      TP      FP      FN
            LOC    0,7901  0,6215  0,6957  734     195     447
           MISC    0,7464  0,4079  0,5275  412     140     598
            ORG    0,8253  0,5633  0,6695  699     148     542
            PER    0,7497  0,4768  0,5829  668     223     733
         Totals    0,7807  0,5200  0,6242  2513    706     2320
```
without
```
CRFClassifier training ... done [895.0 sec].
...
CRFClassifier tagged 51645 words in 3068 documents at 21626,88 words per second.
         Entity    P       R       F1      TP      FP      FN
            LOC    0,7887  0,6257  0,6978  739     198     442
           MISC    0,7408  0,4188  0,5351  423     148     587
            ORG    0,8239  0,5657  0,6708  702     150     539
            PER    0,7486  0,4825  0,5868  676     227     725
         Totals    0,7784  0,5256  0,6275  2540    723     2293
```
The slight variations in quality are probably due to using log1p instead of log(1+) in some places, etc. I'd assume that increasing QNsize back to 10 will make up for that. And with these patches - because they also reduce memory consumption - I can run QNsize 10 on connl with 2 GB limit, while I could not with the unmodified version.

kno10:
Hi, I have rebased my changes on the lastest git version (as #584 is fixed).
I have also reordered them such that JaFaMa is _only_ in the last commit.
If you pick the changes except the JaFaMa change, then you will get all the other optimizations, without introducing the JaFaMa dependency (runtime 840.1).

kno10:
Updated benchmarks (other settings than above):
```
3.8.0          CRFClassifier training ... done [4645.4 sec].
3.9.0-20180131 CRFClassifier training ... done [4616.0 sec].
patch-jafama   CRFClassifier training ... done [3534.3 sec].

3.8.0          CRFClassifier tagged 51645 words in 3068 documents at 15602.72 words per second.
3.9.0-20180131 CRFClassifier tagged 51645 words in 3068 documents at 17808.62 words per second.
patch-jafama   CRFClassifier tagged 51645 words in 3068 documents at 26242.38 words per second.

3.8.0          Totals	0.8386	0.6872	0.7554	3321	639	1512
3.9.0-20180131 Totals	0.8386	0.6872	0.7554	3321	639	1512
patch-jafama   Totals	0.8388	0.6872	0.7555	3321	638	1512
```
I.e. training time reduced by ~23%, or from 1 hour 17 minutes to just under 1 hour.

The gains in prediction appear larger, but that is likely misleading. It reduced the time from 2.9 to 1.9 seconds (so by 30%), but the measurement accuracy is not as good, and I/O may have an impact here. Much of the optimization focuses on reducing the training time, as I am experimenting with some larger training data sets. But maybe the optimized feature extraction and Viterbi do help.

The result quality is almost identical. Because of numerical changes, this version does 1 mistake less on this test data set, but that is not significant.

AngledLuffa:
How do you feel about separating this into two pull requests, one with the new library and one with the rest of the changes?

kno10:
It already is (at least mostly). JaFaMa is only added in one of the last commits, 94518ae82766ceddc0692ce852b4d85229ed7084
I began further restructuring these patches to make this the last commit in the series, but never got around to finishing and pushing that rebase. 
I just force-pushed that rebased version, as it seems to now have the JaFaMa patch last.

kno10:
Breaking this completely into two does not make as much sense:
Some of the earlier patches change / improve / enhance some parts of ArrayMath, for example, where JaFaMa is now used. So they are not completely independent code-wise.
Some of the other patches in this pull (such as the 2x1 character Javadoc patch) are straightforward. "Optimize feature generation" is a separate pull request (as it changes lines that use Math).

AngledLuffa:
Well, I integrated the NERFeatureFactory changes (and the tiny javadoc changes).  The rest of the changes are a lot of changes to code I haven't spent much time with, so I need to go over them in a bit of detail in order to accept them.  Thanks for the contributions!  I will continue working on this.

src/edu/stanford/nlp/math/SloppyMath.java
```diff
@@ -294,13 +294,8 @@ public static double logAdd(double lx, double ly) {
...
       negDiff = lx - ly;
     }
-    if (max == Double.NEGATIVE_INFINITY) {
-      return max;
-    } else if (negDiff < -LOGTOLERANCE) {
-      return max;
-    } else {
-      return max + Math.log(1.0 + Math.exp(negDiff));
-    }
+    return (max == Double.NEGATIVE_INFINITY || negDiff < -LOGTOLERANCE) ? max : //
```
AngledLuffa: Does this have anything to do with efficiency, or just more compact because of ?: operator?  why the // at the end?

src/edu/stanford/nlp/tagger/maxent/Dictionary.java
```diff
@@ -38,8 +38,7 @@ public Dictionary() {
 
   void fillWordTagCounts(Map<String, IntCounter<String>> wordTagCounts) {
     for (Map.Entry<String, IntCounter<String>> wordTagCount : wordTagCounts.entrySet()) {
-      TagCount count = new TagCount(wordTagCount.getValue());
-      dict.put(wordTagCount.getKey(), count);
+      dict.put(wordTagCount.getKey(), new TagCount(wordTagCount.getValue()));
```
AngledLuffa: Again, is this an efficiency issue or a compact code issue?

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
```diff
@@ -357,15 +357,13 @@ private void setHistory(int current, History h, int[] tags) {
...
-        continue;
-      } //but shouldn't happen
-      if (j >= size + left) {
-        break;
-      } //but shouldn't happen
-      h.setTag(j - left, maxentTagger.tags.getTag(tags[j]));
+    // Loop range, with guards:
+    final int s = Math.max(current - left, left);
+    final int e = Math.min(current + right, size + left - 1);
+    for (int j = s; j <= e; j++) {
```
AngledLuffa: Random question from a distant memory.  I recall that in C++, ++i and i++ have slightly different performance because the return value of i++ is different from the final value of i.  Is that true in Java?  I would expect ++j to be a tiny bit faster.

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
```diff
@@ -821,28 +794,30 @@ public double scoreOf(int[] sequence) {
     return getScores(history);
   }
 
-  // todo [cdm 2013]: Tagging could be sped up quite a bit here if we cached int arrays of tags by index, not Strings
+  // todo [cdm 2013]: Tagging could be sped up quite a bit here if we cached int
+  // arrays of tags by index, not Strings
   protected String[] stringTagsAt(int pos) {
-    if ((pos < leftWindow()) || (pos >= size + leftWindow())) {
+    pos -= leftWindow(); // Adjust once, for everyting below:
```
AngledLuffa: typo (everything)

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
```diff
@@ -574,60 +567,39 @@ private void cleanUpScorer() {
...
-      int[] fAssociations = maxentTagger.fAssociations.get(kf).get(val);
-      if (fAssociations != null) {
-        for (int j = 0; j < tags.length; j++) {
-          String tag = tags[j];
-          int tagIndex = maxentTagger.tags.getIndex(tag);
-          int fNum = fAssociations[tagIndex];
-          if (fNum > -1) {
-            double score = maxentTagger.getLambdaSolve().lambda[fNum];
-            if (VERBOSE) {
-              fullScores[kf][j] = score;
```
AngledLuffa: Can we keep the debug statements?  They didn't make it into the refactored method below.

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
```diff
@@ -574,60 +567,39 @@ private void cleanUpScorer() {
     return scores;
   }
 
-  // todo [cdm 2016]: Could this be sped up a bit by caching lambda array, extracting method for shared code?
-  // todo [cdm 2016]: Also it's allocating java.util.ArrayList$Itr for for loop - why can't it just random access array?
   /** Returns an unnormalized score (in log space) for each tag. */
   private double[] getApproximateHistories(String[] tags, History h, List<Pair<Integer,Extractor>> extractors, List<Pair<Integer,Extractor>> extractorsRare) {
```
AngledLuffa: Any equivalent optimizations to be done in getExactHistories?

AngledLuffa:
Left a few comments on the first actual change, the one touching a bunch of
tagger code.  Mostly seems good.  Thanks for the help!

src/edu/stanford/nlp/math/SloppyMath.java
kno10: The // is to prevent code formatters from making this a single line. This is my usual code style with ternary statements to make them more readable.

Ternary statements, while logically being the same and frowned upon by some purists, tend to be faster in my experiments with Java. This may be because of two factors:
1. Smaller byte code size (58 vs. 62)
2. Single return, which may make inlining easier, too.

src/edu/stanford/nlp/tagger/maxent/Dictionary.java
kno10: Similar to above. The Java compiler does a quite literal translation, so the resulting bytecode with the local variable is usually about 2 bytes more - an astore and an aload to a local register. This will be optimized away by the JIT compiler easily, but may affect the decision when to inline a function (which depends, to my understanding, a lot on bytecode size - functions with fewer than 12 are always inlined unless multimorphic, and when code is considered to be hot also larger functions will be inlined eventually).

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
kno10: Indeed ++i is supposed to be faster, unless the compiler manages to optimize this. Here I would bet that Java can optimize this away - and so should any decent C/C++ compiler.
In C++ it may make a difference if operator++ is a pretty complex function - but in Java, this only exists for primitives, hence it should not matter.
According to the answers to this post:
https://stackoverflow.com/questions/12396740/is-x-more-efficient-than-x-in-java
both will both produce an iinc bytecode; just potentially in a different order. Hence, it should produce the exact same byte code here and then cannot make a difference.

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
kno10: They were added to the codebase after my initial version, which is how they got lost. Looked like dead code when rebasing. I'll add them back.

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
kno10: Yes. Maybe it could be further unified; it seems the exact is going over all, the approximate only over a subset of tags.

kno10:
Pushed a new version that also refactors exact histories in a similar way.

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
AngledLuffa: Neat, thanks!

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
AngledLuffa: Thanks!  Maybe I'll change the names to runExactExtractor and runApproximateExtractor

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
```diff
@@ -821,28 +761,30 @@ public double scoreOf(int[] sequence) {
...
-      return arr1;
+    if (originalTags != null) {
+      final String orig = originalTags.get(pos);
+      if (orig != null) {
+        return new String[] { orig };
+      }
     }
 
+    String[] arr1;
     String word = sent.get(pos - leftWindow());
```
AngledLuffa: missed a leftWindow() here :/

AngledLuffa:
How do you feel about rebasing now that a few of the changes have been merged?  Also, if you could push the JaFaMa change as far down as possible, that would make the process of merging the rest of the changes easier.  Thanks for contributing.

src/edu/stanford/nlp/tagger/maxent/TestSentence.java
kno10: well spotted. I may have introduced this just yesterday with the rebases, because this fragment had been independently improved (I recall having had a compile error because of this arr1).

AngledLuffa:
It was easy to spot when I tested the tagger :)

kno10:
Rebased once more.
Github does not show the patches in the correct order. The jafama patch was the last in the series the entire time. I guess github is confused because they are not in temporal order; during rebasing I moved some of the simpler patches first, and the larger such as JaFaMa last. Right now, it is also showing your last reply inbetween of the patches...

src/edu/stanford/nlp/tagger/maxent/TTags.java
```diff
@@ -49,6 +49,7 @@
    *  closedTagThreshold will be considered closed.
    */
   private boolean learnClosedTags = false;
+  private String[] openTagsCache;
```
AngledLuffa: look for openTagsArr on line 31

src/edu/stanford/nlp/optimization/QNMinimizer.java
```diff
@@ -328,7 +328,6 @@ public MaxEvaluationsExceeded(String s) {
     private int maxSize = 100; // This will control the number of func values /
     // gradients to retain.
     private Function mon = null;
-    private boolean quiet = false;
```
AngledLuffa: is this is a case of things having diverged from the time you first wrote it?  below, all the debug messages are calls to sayln()

AngledLuffa:
Would you check that everything has been merged from "Optimize: avoid reconstructing the same set and array again and again."

src/edu/stanford/nlp/optimization/QNMinimizer.java
kno10: No, this IMHO is even a subtle bug. There are two `quiet` accessible in this context.
https://github.com/stanfordnlp/CoreNLP/blob/eb43d5d9150de97f8061fa06b838f1d021586789/src/edu/stanford/nlp/optimization/QNMinimizer.java#L103
https://github.com/stanfordnlp/CoreNLP/blob/eb43d5d9150de97f8061fa06b838f1d021586789/src/edu/stanford/nlp/optimization/QNMinimizer.java#L331
My patch intentionally removes `sayln` because that is an anti-pattern in logging. As is, it would first construct the log message, then call sayln, and only then notice that it is silenced. The "quiet" check should be outside; redundant checks can then be merged.

kno10:
The main conflict when rebasing was with 4f459f76dce08cc56e293646e0146988ce89b895 and 8da1176c52d37eb2d3a4a3b2323e90ce6cf69db7 in TestSentence.java - that is where the open tags are used. I think that commit is then resolved.

AngledLuffa:
Oof, alright, I'll take some time to sort through that with the help of
your change.  Thanks for pointing that out.

AngledLuffa:
Mind confirming that this has been merged?  ```Optimize: avoid reconstructing the same set and array again and again.```

src/edu/stanford/nlp/ie/crf/CRFLogConditionalObjectiveFunction.java
```diff
@@ -776,39 +789,11 @@ public int domainDimension() {
...
-        log.info("weights: " + Arrays.toString(weights));
-        log.info("newWeights["+i+"]: " + Arrays.toString(newWeights[i]));
-        throw new RuntimeException(ex);
-      }
-      index += labelSize;
-    }
-    return newWeights;
-  }
-
-  public double[][] to2D(double[] weights) {
```
AngledLuffa: This is actually used by a subclass of CRFLogConditionalObjectiveFunction which we have hidden in our internal codebase.  Something we didn't want to publicly release because of licensing issues around some of the other libraries it uses.  

Any particular reason for removing it?  Just general cleanliness?

Actually we could easily move it to the subclass I think, so not really a problem

AngledLuffa:
also, i integrated a few of the earlier commits, so if you could rebase on top of that it would be great for making the rest of the integration simpler.

AngledLuffa:
(and drop the commits which were integrated, i suppose)

src/edu/stanford/nlp/optimization/QNMinimizer.java
```diff
@@ -522,45 +521,55 @@ private eState toContinue(StringBuilder sb) {
...
-      return s.get(ind);
+      return s[ind];
     }
 
     double[] getY(int ind) {
-      return y.get(ind);
+      return y[ind];
+    }
+
+    void removeFirst() {
```
AngledLuffa: Do you have any interest in doing the circular buffer?  Should I take a look at it?

src/edu/stanford/nlp/optimization/QNMinimizer.java
```diff
@@ -582,9 +591,10 @@ void free() {
     }
 
     void clear() {
-      s.clear();
-      y.clear();
-      rho.clear();
+      used = 0;
```
AngledLuffa: fill with null to deference the arrays?  probably worth a comment

src/edu/stanford/nlp/optimization/QNMinimizer.java
AngledLuffa: I had missed the first quiet.  Thanks for pointing that out.

src/edu/stanford/nlp/optimization/QNMinimizer.java
AngledLuffa: i added this comment myself

src/edu/stanford/nlp/ie/crf/CRFLogConditionalObjectiveFunction.java
AngledLuffa: I realized the other class was actually using this as a library, so I left the original here (with similar changes to the ones you made).

src/edu/stanford/nlp/math/SloppyMath.java
AngledLuffa: Interesting.  Thanks for the explanation

src/edu/stanford/nlp/optimization/QNMinimizer.java
AngledLuffa: Actually, even among micro-optimizations this would be the most micro of optimizations, right?  It's basically copying 20 references per iteration, which would be the tiniest drop in the bucket of the processing time for model training

src/edu/stanford/nlp/optimization/QNMinimizer.java
```diff
@@ -522,45 +521,55 @@ private eState toContinue(StringBuilder sb) {
     // Diagonal Options
     // Line search Options
     // Memory stuff
-    private List<double[]> s = null;
-    private List<double[]> y = null;
-    private List<Double> rho = null;
+    private double[][] s;
```
AngledLuffa: i put back the `= null` for consistency

src/edu/stanford/nlp/optimization/QNMinimizer.java
kno10: I am quite busy right now, so I can't reply to all comments, and much of the edits I did were like 1.5-2 years ago.
A circular buffer may be more elegant: instead of having a `used` that will at some point always be 19 and 20, you would have a counter pointing to the next element to be replaced; and if you need shrinking at all, one could overwrite values with `null`. Right now, what is the meaning of the position of a value in this arrays except recency? Once filled, the current code removes s[0], y[0]; moves everything up front, and then reinserts s and y at the end again (because the code now recycles the arrays; it most likely did not do so originally). A simple "latest=-1" variable should suffice, and when you do a new solution, you increase latest (wrapping around to 0) and work on s[latest] and y[latest]. So it would likely result in simpler code.

src/edu/stanford/nlp/ie/crf/CRFClassifier.java
```diff
@@ -959,42 +917,26 @@ protected void makeAnswerArraysAndTagIndex(Collection<List<IN>> ob) {
...
-    // for (int j = 0; j < FeatureFactory.win[i].length; j++) {
-    // featuresC.addAll(featureFactory.features(info, loc,
-    // FeatureFactory.win[i][j]));
-    // }
-    // features.add(featuresC);
-    // }
+    ArrayList<Collection<String>> features = new ArrayList<>(windowSize);
+    List<double[]> featureVals = flags.useEmbedding ? new ArrayList<>(1) : null;
 
-    // todo [cdm Aug 2012]: Since getCliques returns all cliques within its bounds, can't the for loop here be eliminated? But my first attempt to removed failed to produce identical results....
```
AngledLuffa: Any reason for dropping this comment?  The loop is still here, just pushed into eachClique

kno10: I think the comment refers to the outer loop, over `i`.
And even when pushing this loop into getCliques, it would still be a for loop there, wouldn't it?
At some point you'll want all features at all positions. So I concluded it cannot be removed.

kno10:
Pushed a rebased version, only three patches remaining.

AngledLuffa:
Excellent, thanks for the repush

AngledLuffa:
"use subclasses instead of switch" does not seem to speed things up for me.  I'm comparing with and without this change on a AMD Ryzen Threadripper 2920X

AngledLuffa:
Maybe I spoke too soon.  Running it again seems to show some speedup.  Don't think anyone has access to my machine...

kno10:
The impacts of most changes will be huge; but they may surface in particular cases only or just accumulate. The JaFaMa patch is the one with the largest impact; and all that reduce memory consumption (such as the feature generator improvement) will also be much more measurable.
Most likely Hotspot will manage to optimize the switch okay; but it will certainly optimize the subclass well (not even compiling the other class). The resulting smaller methods may also give it a larger optimization horizon, but they will likely still be "big methods" and not inlined anywhere else. So I would not expect this commit to make a big difference, you could omit it if you don't like the OOP it introduces.

AngledLuffa:
Well, even for the subclassing one, 1% is 1%

AngledLuffa:
Well, there's two changes left, "Unnecessary copying of clique lists" and the JaFaMa change.  

I've rerun several end to end trainings with the clique list change, and on average it seems to make training an NER model slightly slower.  I'm interested in hearing about your results if they are different, though.  

The JaFaMa change introduces a new library, and we're a little skeptical about adding more dependencies.  Hopefully it will be easier to maintain locally for yourself now that most of the rest of the changes have been merged.

Overall I think we've integrated everything we can and will close the pull request.  Any thoughts?

AngledLuffa:
Actually, I was a bit tunnel visioned on the previous few changes affected training time and not test time.  Let me run a few speed tests on test time before declaring the clique list change unhelpful

kno10:
The "unnecessary copying" patch should not cause a slowdown; but it should reduce the garbage collector load. So the benefits will likely show with large benchmarks, and when working close to the minimum memory requirements. It may be worth using A) serialGC and locking it to a single core for benchmarking, to ensure the GC load is not just offloaded in background to the other CPUs, and B) monitoring garbage collector statistics.
For JaFaMa, if you do not want to introduce an external dependency, you can just copy the classes to include it as source (this is a bit hacky though, as others may then have two copies of this on their classpath). This is a very compact library (<10 classes), doing little more than a faster java.lang.Math. As such, it also does not change much (rare releases, as the functionality is stable), so it won't need constant updating as with many other libraries (that then may bring incompatible changes). I am extremely restrictive with external dependencies on my own projects (also because I have seen pretty bad code even in highly popular Apache libraries), but JaFaMa looked solid to me and made a difference. Maybe you can have a look at the benchmark results first - this patch probably makes the largest difference - then decide.

AngledLuffa:
I find that even my lazy test of running the official NER test 10 times in
the same process shows an improvement at test time.  So I take back my
earlier comment about "unnecessary copying", since it noticeably helps at
test time compared to any possible change at train time.

I'll do as you suggest and benchmark the JaFaMa changes before making any
decisions (maybe @maning will have an opinion after seeing the results)

Thanks again for all the help.

AngledLuffa:
@manning not whoever it was I accidentally pinged there

src/edu/stanford/nlp/stats/Counters.java
```diff
@@ -1674,7 +1675,7 @@ private Counters() {} // only static methods
     C result = (C) c.getFactory().create();
     for (E key : c.keySet()) {
       double count = c.getCount(key);
-      double noise = -Math.log(1.0 - random.nextDouble()); // inverse of CDF for
+      double noise = -FastMath.log(1.0 - random.nextDouble()); // inverse of CDF for
```
AngledLuffa: not sure if this is in any way critical enough to need a log1p as well

JavaNLP-core.eml
```diff
@@ -86,6 +86,11 @@
 		<relative-module-src project-related="jar://$PROJECT_DIR$/projects/core/libsrc/ejml-0.23-src.zip!/"/>
 		<relative-module-cls project-related="jar://$PROJECT_DIR$/projects/core/lib/ejml-0.23.jar!/"/>
 	</lib>
+	<lib name="jafama-2.3.0.jar" scope="COMPILE">
```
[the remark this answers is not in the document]
kno10: Yes. I'm usually using 2.3.1 now. But for some time, it was not available via maven.

src/edu/stanford/nlp/classify/Dataset.java
```diff
@@ -658,7 +661,7 @@ public void selectFeatures(int numFeatures, double[] scores) {
     for (int i = 0; i < labelIndex.size(); i++) {
       double labelCount = labelCounter.getCount(labelIndex.get(i));
       double p = labelCount / size();
-      entropy -= p * (Math.log(p) / Math.log(2));
+      entropy -= p * FastMath.log(p) * LN_TO_LOG2;
```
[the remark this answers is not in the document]
kno10: Definitely, since multiplications are usually faster than divisions.

src/edu/stanford/nlp/parser/lexparser/EvaluateTreebank.java
```diff
@@ -391,7 +392,7 @@ else if(pwFileOut != null) {
           double denom = ArrayMath.logSum(logScores);
           for (double logScore : logScores) {
             double logPr = logScore - denom;
-            entropy += Math.exp(logPr) * (logPr / Math.log(2));
+            entropy += FastMath.exp(logPr) * (logPr / FastMath.log(2));
```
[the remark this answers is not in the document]
kno10: It should likely use the same constant `LN_TO_LOG2=1/log(2)` as above, with multiplication rather than division. The numerical differences between division and multiplication with the inverse should be negligible here, since `log(2)` does not have an exact representation in floating point.

src/edu/stanford/nlp/parser/lexparser/ExhaustiveDependencyParser.java
```diff
@@ -637,7 +638,7 @@ public void displayHeadScores() {
             float penalty = 0.0f;
             if (bd != 0) {
               penalty = (float) dg.score(words[hWord], hTag, -2, -2, aWord > hWord, 0);
-              penalty = (float) Math.log(1.0 - Math.exp(penalty));
+              penalty = (float) FastMath.log(1.0 - FastMath.exp(penalty));
```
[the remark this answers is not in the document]
kno10: For `log1p` I am not entirely certain about the numerics for x<0. You should get some more expert opinion on this, I guess. My understanding of the issue is that if the exp is much smaller than 1, the subtraction is lossy; and hence `log1p` should indeed be more accurate. But I don't know of the `log1p` function is only intended/optimized for x>0. So I didn't want to just change this (although I do think its better).

src/edu/stanford/nlp/parser/lexparser/SplittingGrammarExtractor.java
```diff
@@ -409,8 +410,8 @@ public void splitBetas() {
           for (int i = 0; i < parentStates; ++i) {
             for (int j = 0; j < childStates; ++j) {
               double childWeight = 0.45 + random.nextDouble() * 0.1;
-              newBetas[i][j * 2] = betas[i][j] + Math.log(childWeight);
-              newBetas[i][j * 2 + 1] = betas[i][j] + Math.log(1.0 - childWeight);
+              newBetas[i][j * 2] = betas[i][j] + FastMath.log(childWeight);
+              newBetas[i][j * 2 + 1] = betas[i][j] + FastMath.log(1.0 - childWeight);
```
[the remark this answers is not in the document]
kno10: Same as above, I would assume - but I do not know - that `log1p` improves precision here.

src/edu/stanford/nlp/stats/Counters.java
kno10: Here the better solution is actually `-Math.log(random.nextDouble())`. Since the random values are supposedly uniform, it does not make sense to use 1- at all.

AngledLuffa: random.nextDouble() can produce 0, right?  Perhaps that was the consideration.

AngledLuffa:
It looks like using JaFaMa speeds things up by 7-8%.  In discussing things with @manning, I think we've concluded that we'd prefer to not have the extra dependency with any future pitfalls it might introduce.  On the bright side, we've integrated everything else and definitely appreciate the changes.  Any thoughts on this?

kno10:
Ok. Thank your for your big effort in merging!
Its fine that you opted for sticking with standard Java `Math`; for my ELKI project I also try to keep external dependencies to a minimum.

src/edu/stanford/nlp/math/ArrayMath.java
```diff
@@ -1455,7 +1458,7 @@ public static double klDivergence(double[] from, double[] to) {
       double num = from[i] / tot;
       double num2 = to[i] / tot2;
       // System.out.println("num is " + num + " num2 is " + num2);
-      kl += num * (Math.log(num / num2) / Math.log(2.0));
+      kl += num * (FastMath.log(num / num2) / FastMath.log(2.0));
```
AngledLuffa: Here's another log(2) fwiw

AngledLuffa:
Closing on account of all pieces except JaFaMa merged (including the Log2 changes from the final change)

kno10:
@AngledLuffa btw, do you have some benchmark results before/after merging my other changes (probably a benchmark 3.9.2 vs. snapshot would be informative already=?

Likely the differences are different from training and labeling; and I'd imaging that some savings show up as reduced garbage collector activity; which may help long running deployments such as web app more than a regular benchmark.

AngledLuffa:
*facepalm*

Actually I just individually tested the changes and didn't test them as an
aggregate.  Another confounding factor is that I had separately fixed the
caching of tag lists before picking up this changelist.  If you're curious
about exactly how much effect you had, wanted to put this down as a resume
line item, or whatever, you can probably either compare 3.9.2 against the
last git commit with your name on it if you want to include the tag lists,
or you can compare the first git commit with your name on it against the
last.  I tried to remember to credit you for each of the changes I
integrated.

kno10:
Just out of curiosity; so it is okay if it includes the caching issue, too. To get a feeling of how much performance changed; and how this related to the extra 6-8% with the JaFaMa patch. In my single experiment, the total difference was over 20% faster (in training), so if we have 15% without JaFaMa that would already be quite good. When I started this work, it meant I could reduce a multi-day runtime by a few hours, so even a few % would matter.

AngledLuffa:
Just uncovered a small bug in an unused codepath.  In CRFClassifier:allLabels, the lines

        label[j]++;
        if (label[j] >= numClasses) { do stuff }

were replaced with

        if (label[j]++ < numClasses) break;

but the ++ is not saved to the value until after the conditional is evaluated, right?

kno10:
Yes, these should be `++label[j]` to increment first.

PR #473, opened 2017-06-30
Add unique matches option for CoreNLPServer Semgrex

dan-zheng:
## Description

A "unique matches" option has been added for Semgrex queries on a CoreNLP Server.

When `unique` is set to true (and `filter` is set to false), the list of matches returned by the server contains only `unique` entries. This is useful because Semgrex users are oftentimes only interested in unique matches.

`SemgrexMatcher` already contains the `findNextMatchingNode()` function for finding the next unique match, so the `unique` option simply calls that function instead of the default `find()`.

## Example

**POST request body**
```
I ran.
```

**POST request query (without `unique`)**
```
Query: "http://localhost:9000/semgrex?pattern={} [< {} | > {}]"

Response:
{
    "sentences": [
        {
            "0": {
                "text": "ran",
                "begin": 1,
                "end": 2
            },
            "1": {
                "text": "ran",
                "begin": 1,
                "end": 2
            },
            "2": {
                "text": ".",
                "begin": 2,
                "end": 3
            },
            "3": {
                "text": "I",
                "begin": 0,
                "end": 1
            },
            "length": 4
        }
    ]
}
```

**POST request query (with `unique`)**
```
Query: "http://localhost:9000/semgrex?pattern={} [< {} | > {}]&unique=true"

Response:
{
    "sentences": [
        {
            "0": {
                "text": "ran",
                "begin": 1,
                "end": 2
            },
            "1": {
                "text": ".",
                "begin": 2,
                "end": 3
            },
            "2": {
                "text": "I",
                "begin": 0,
                "end": 1
            },
            "length": 3
        }
    ]
}
```

### DIFF
--- a/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
+++ b/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
@@ -977,10 +977,13 @@ public class StanfordCoreNLPServer implements Runnable {
           // (get whether to filter / find)
           String filterStr = params.getOrDefault("filter", "false");
           final boolean filter = filterStr.trim().isEmpty() || "true".equalsIgnoreCase(filterStr.toLowerCase());
+          // (in case of find, get whether to only keep unique matches)
+          String uniqueStr = params.getOrDefault("unique", "false");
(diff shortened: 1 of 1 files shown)


manning:
Thanks! Merging.

Issue #472, opened 2017-06-30
NER and TokensRegex not working together in StanfordCoreNLPServer 3.8.0


nkrot:
(copying over my question from StackOverflow in the hope to get a quicker answer)

In brief: I have observed that `StanfordCoreNLP` and `StanfordCoreNLPServer` (v.3.8.0) work differently when annotators include both `ner` and `tokensregex`. More specifically, in the server mode results of `tokensregex` and `ner` are not combined and only results of `ner` are outputted.

Now description at length:

The server `StanfordCoreNLPServer` is started with `-serverProperties` option pointing to the file with the following content:

    annotators = tokenize,ssplit,pos,lemma,ner,tokensregex

    pos.model = edu/stanford/nlp/models/pos-tagger/german/german-hgc.tagger

    tokenize.language = de
    tokenize.options = normalizeCurrency=false

    ner.model = edu/stanford/nlp/models/ner/german.conll.hgc_175m_600.crf.ser.gz
    ner.applyNumericClassifiers = false
    ner.useSUTime = false

    customAnnotatorClass.tokensregex = edu.stanford.nlp.pipeline.TokensRegexAnnotator
    tokensregex.rules = deu/ner.money.tre.txt

The rule is deliberately made super simple and dumb. There is nothing else but these lines in the file `deu/ner.money.tre.txt`:

    ne = { type: "CLASS", value: "edu.stanford.nlp.ling.CoreAnnotations$NamedEntityTagAnnotation" }

    { ruleType: "tokens",
      pattern: ( /Millionen/ ),
      action: ( Annotate($0, ne, "HUGE") ) }

The text snippet to tag is as follows:

     mit einem Gesamtwert von mehr als € 5 Millionen in Frankfurt am Main .

Experiment 1. Console `StanfordCoreNLP` works as expected outputting both `ner` and `tokensregex` tags. The option `-props` points to the file with properties as described above.

    java -cp "*:/path/to/models-2017-06-09/*" -Xmx2g edu.stanford.nlp.pipeline.StanfordCoreNLP -props ./deu/StanfordCoreNLP-deu.properties -file text.in -outputFormat json

grepped output contains all correct NEs

    ...
    { "word": "Millionen", "ner": "HUGE" },
    ...
    { "word": "Frankfurt", "ner": "I-LOC" },
    { "word": "am",        "ner": "I-LOC" },
    { "word": "Main",      "ner": "I-LOC" },
    ...

[...]

PR #519, opened 2017-09-09
StanfordCoreNLPServer: adding properties parametrization to /tokensregex, /semgrex, and /tregex

gerardobort:
_This PR may probably respond and fix the issue described here: https://github.com/stanfordnlp/CoreNLP/issues/472_

## Issue Description
`StanfordCoreNLPServer` has 4 endpoints:

1. `/` 👌
2. `/tokensregex`
3. `/semgrex`
4. `/tregex`

The first one performs the annotations over the given text, receiving `properties` and `pipelineLanguage` via query-strings.  The other three are labeled as "CoreNLP Tools", and works differently: they only accept `filter` and `pattern`.

The *problem* comes up when you are not working with the default language (English). In this scenario the last 3 enumerated endpoints won't accept any other than English.  This results in a separate annotation pipeline initialized for the "Tools" (an English pipeline), making the POS matches invalid for the three utilities.

## Solution
This change moves the `getProperties` private method from the first endpoint handler class to the container class, in order to be reused by the other endpoint handlers too.  Also there's some minor tweaks on the JS files that handles the "Tools" form submissions.

I tested this with Spanish, and after these changes it works pretty well, either using Universal Dependencies or the "distsim" tagger.  Please, note that the changes are pretty superficial.

### Before
Note that Semgrex matches `FW` for "Yo", instead of `pp0000000` as it appears in POS.
screen shot 2017-09-09 at 9 54 01 pm

### After
Apart of me using UD, note that Semgrex matches correctly the same POS tags as above.
screen shot 2017-09-09 at 9 53 59 pm

### DIFF
--- a/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
+++ b/src/edu/stanford/nlp/pipeline/StanfordCoreNLPServer.java
@@ -340,6 +340,89 @@ public class StanfordCoreNLPServer implements Runnable {
     return impl;
   }
 
+  /**
+   * Parse the parameters of a connection into a CoreNLP properties file that can be passed into
+   * {@link StanfordCoreNLP}, and used in the I/O stages.
+   *
+   * @param httpExchange The http exchange; effectively, the request information.
+   *
+   * @return A {@link Properties} object corresponding to a combination of default and passed properties.
+   *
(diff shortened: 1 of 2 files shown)


manning:
Thanks so much, this looks really useful. Will merge it.

PR #527, opened 2017-09-21
Refactoring some opportunities to use lambda expressions

rbonifacio:


manning:
Thanks for the suggestions, but ... while this is an issue that people split on, I'm just not convinced that it is a good idea to refactor every little loop into a lambda because you can. See e.g., the discussion at: https://blog.jooq.org/2015/12/08/3-reasons-why-you-shouldnt-replace-your-for-loops-by-stream-foreach/ . I think forEach is fine to use if it makes things simpler or faster, but for simple cases, often a for-loop seems easier to read to me. Maybe I'm just getting old.

rbonifacio:
@manning Sure, I agree with you. We have been implementing some refactoring tools to assist developers to use new constructs of the Java language. As part of this research, we found that transforming for-loops to lambda is a kind of transformation that developers do not receive well. Perhaps due to the issues cited on the blog post you shared. I will close this pull request. Thanks.

PR #533, opened 2017-09-27
Unicode case insensitive matching

diadistis:
This PR fixes the ignore case option in several places for non US-ASCII texts.


manning:
Thanks!

PR #543, opened 2017-10-15
(refactor) some opportunities to use Switch String

desmorto:


manning:
Thanks, will merge. But it seems like there are a couple of opportunities to do more which the automated tool can't handle. E.g., in `src/edu/stanford/nlp/coref/data/Mention.java`, it doesn't handle when two cases have been combined in one `if` via `||`....

PR #544, opened 2017-10-15
(refactor) some opportunities to use Anonymous inner Classes

desmorto:


manning:
Thanks! Mainly applied.

But it's not useful to refactor CoreNLPProtos.java, since it's autogenerated from the Google protobuf compiler.

PR #545, opened 2017-10-15
(refactor) some opportunities to use lambda functions

desmorto:


manning:
Hi desmorto, thanks for trying to help, but I'm not going to merge this change. I'm not against the use of Java 8 streams and functional constructs, but I'm afraid that I'm not convinced that it is a good idea to bulk change tons of conventional for-loops to streams. In simple cases, I think many people still find conventional for-loops easier to read, and, in these examples, most of the underlying collections are small, not large, and in general in an unthreaded context, you most frequently get performance losses from changing small for-loops to streams. These issues have been documented in various places, including:

https://blog.jooq.org/2015/12/08/3-reasons-why-you-shouldnt-replace-your-for-loops-by-stream-foreach/
https://jaxenter.com/java-performance-tutorial-how-fast-are-the-java-8-streams-118830.html
https://blog.takipi.com/benchmark-how-java-8-lambdas-and-streams-can-make-your-code-5-times-slower/
https://blog.codefx.org/java/stream-performance/

PR #546, opened 2017-10-15
(refactor) some opportunities to use Variadic Arguments

desmorto:


manning:
Thanks for the suggestions, but, sorry, we're not going to merge these changes. We'd be happy to add varargs to any appropriate method where it seems like it would provide a useful API improvement, but it doesn't seem useful to change every method that takes a final array to a method that takes varargs....   See perhaps the last paragraph of this page: https://docs.oracle.com/javase/8/docs/technotes/guides/language/varargs.html

PR #577, opened 2017-11-26
refactoring using lambda expressions

waltim:


manning:
Thanks for the suggestions. We're not against functional programming techniques, but we're not going to be changing things to use streams except where it seems clearly a better way to do things. So, sorry, I'm going to close this. See #527 and #545 for more discussion.

PR #588, opened 2017-12-14
Don't swallow Errors in MulticoreWrapper

kno10:
If an `Error` (e.g., `OutOfMemoryError`, `ClassNotFoundError`) is thrown by a multithreaded job, the thread is never returned to the pool, and the `MulticoreWrapper` eventually hangs infinitely.

This change will A) catch `Errors` and print the stack trace, B) always calls the callback, to return to the thread pool.

I think it would be better to somehow bubble the exception/error to the caller. But this is the minimal change to at least see the error message printed and allow the caller to fail (when seeing a null value).


manning:
Thanks a lot! Merging.

PR #589, opened 2017-12-22
Update GrammaticalStructureConversionUtils.java

abinj:
Fixed the issue related to GrammaticalStructure related root and label funtion call.


manning:
Thanks. Merging and doing a bit more to make these variables private.

PR #592, opened 2018-01-03
Remove defect "covariance" function from ArrayMath

kno10:
Since the function is unused, I suggest to just drop it for now.
It was invalid. The covariances do not involve the square root, and because the values can be negative, this would frequently yield NaN values.
The current implementation was also not very fast, because of nested array accesses.


manning:
Seems fair, thanks!

PR #593, opened 2018-01-03
Avoid shrinking qn m size to 0 on out of memory.

kno10:
This could cause an infinite loop, qn.mem must remain at least 1. We can't work with 0 candidates.


manning:
Thanks! Merging.

PR #598, opened 2018-01-11
Add BIAS feature only once

kno10:
I think this should be outside of the loop.


manning:
Yeah, it does seem like the bias feature should only be added once….

PR #612, opened 2018-01-25
Simplify code, don't swallow errors (but log them)

kno10:
As the patterns are only used once, there is no benefit of using Pattern.compile, Matcher over String.replaceAll.

The "catch(Exception e) {}" obviously is a bad idea, as it silently ignores any error. I have replaced it with a logged catch(IOException).

Using a larger string buffer reduces the memory reallocations.

However, the code should be further cleaned up:

1. Doing all the work in the constructor is usually a bad idea, in particular wrt. subclassing.
2. On error, it should probably *throw* an IOException, not catch it (but exception in constructors), but this will require changes in many places for proper IO error handling. It would also be possible to rethrow this as an unchecked RuntimeException.
3. `"^" + delimiter" is not guaranteed to produce the desired regular expression. For example one could use the pattern `a|b`, which will then remove all `b`s instead of just leading `b`s. So the code *is* buggy. Instead, I would suggest to just split with the delimiter, check if the first and/or last elements are empty, then use `subList`.

### DIFF
--- a/src/edu/stanford/nlp/objectbank/DelimitRegExIterator.java
+++ b/src/edu/stanford/nlp/objectbank/DelimitRegExIterator.java
@@ -2,14 +2,14 @@ package edu.stanford.nlp.objectbank;
 
 import java.util.function.Function;
 import edu.stanford.nlp.util.AbstractIterator;
+import edu.stanford.nlp.util.logging.Redwood;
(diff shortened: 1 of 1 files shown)


manning:
Thanks a lot!

PR #613, opened 2018-01-25
Don't compile regexp for every line of the lexicon

kno10:
Once is enough. Puts less load on the garbage collector.


manning:
Thank you for your continued contributions!

PR #614, opened 2018-01-26
Optimize feature generation 

kno10:
This reduces the repeated string concatenations when generating NER features.

Previously, for each clique, we would build a set of strings,
then postfix each with the clique (which means creating another string).

The new helper class handles (most) of the concatenations with a shared buffer,
thus reducing the number of temporary string copies substantially,
which means much less work for the garbage collector, and thus higher performance.

(No, JVM would probably not optimize away all the interim strings. They 'escape'
scope via the collections, so escape analysis will not help.)

Note: this also includes #613

### DIFF
--- a/src/edu/stanford/nlp/ie/NERFeatureFactory.java
+++ b/src/edu/stanford/nlp/ie/NERFeatureFactory.java
@@ -32,7 +32,6 @@ import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


AngledLuffa:
Going through some old pull requests...  this seems unlikely to be what makes the NER slow, but do you have evidence otherwise?

kno10:
I do not have exact numbers. This is just one of multiple measures I did to speed up CoreNLP for crunching faster through all of Wikipedia.

Memory is usually key, in particular avoiding Garbage collection. And the feature generation code runs for every token several times.

Seemingly harmless lines such as
https://github.com/stanfordnlp/CoreNLP/blob/8aa8df3de4856a2fe52fe57476135bba4181d80c/src/edu/stanford/nlp/ie/NERFeatureFactory.java#L835-L848
can cause quite some unexpected garbage collection work. Because each of these string concatenation operations usually is translated to something like (1) creating a `StringBuilder` (maybe also a `StringJoiner`), (2) appending the first string, (3) appending the second string, frequently involving a resize (copy) of the character buffer, (4) copying the buffer to a new String object (at which point new Java may consider to convert it from a char to a byte representation); so that usually means creating two objects and 2-3 arrays to store the character data. In certain cases (with a "domain" set) this entire process is then repeated twice. This puts quite some load on the garbage collector.
Calls to `addAllInterningAndSuffixing` will then even involve additional calls to build strings:
https://github.com/stanfordnlp/CoreNLP/blob/eb43d5d9150de97f8061fa06b838f1d021586789/src/edu/stanford/nlp/sequences/FeatureFactory.java#L93-L108

In the proposed patch, the `StringBuilder` as well as its internal arrays only exists once, and they are reused. As this runs many times per word, the impact should be measurable, even in microbenchmarks that only tokenize a short sentence where GC may not yet be an issue. Reducing object allocations and avoding the `+` operator on Strings are some of the most successful Java optimization techniques.

src/edu/stanford/nlp/ie/NERFeatureFactory.java
```diff
@@ -788,125 +899,112 @@ private void generateSlashHyphenFeatures(String word, Collection<String> feature
...
+        if (!lem.isEmpty()) {
+          out.addConcat(lem, "-LEM");
         }
       }
       if (flags.usePrevNextLemmas) {
         String plem = p.getString(CoreAnnotations.LemmaAnnotation.class);
         String nlem = n.getString(CoreAnnotations.LemmaAnnotation.class);
-        if (! "".equals(plem)) {
-          featuresC.add(plem + "-PLEM");
+        if (plem.isEmpty()) {
```
AngledLuffa: Should be

!plem.isEmpty()

?

src/edu/stanford/nlp/ie/NERFeatureFactory.java
```diff
@@ -788,125 +899,112 @@ private void generateSlashHyphenFeatures(String word, Collection<String> feature
...
+          out.addConcat(plem, "-PLEM");
         }
-        if (! "".equals(nlem)) {
-          featuresC.add(nlem + "-NLEM");
+        if (nlem.isEmpty()) {
```
AngledLuffa: Again looks like ! is missing

src/edu/stanford/nlp/ie/NERFeatureFactory.java
```diff
@@ -1579,85 +1670,84 @@ else if (flags.useTitle2) {
...
-          for (Object ob: (Collection)c.get(key)) {
-            featuresC.add(ob + "-" + CoreLabel.genericValues.get(key));
+        Object col = c.get((Class<CoreAnnotation<Object>>) key);
+        if (col instanceof Collection) {
+          for (Object ob: (Collection<?>)col) {
+            out.addConcat(ob.toString(), "-", CoreLabel.genericValues.get(key));
           }
         } else {
-          featuresC.add(c.get(key) + "-" + CoreLabel.genericValues.get(key));
+          out.addConcat(col.toString(), "-", CoreLabel.genericValues.get(key));
```
AngledLuffa: Is there a bug here, originally from the old version?  If you have a null value, the old version would add the feature null-key and the new version will NPE

src/edu/stanford/nlp/ie/NERFeatureFactory.java
```diff
@@ -1747,78 +1833,74 @@ else if (flags.useTitle2) {
...
             if ( ! flags.strictGoodCoNLL) {     // These ones were added later after goodCoNLL
-              featuresCpC.add(pShape + "-PSEQpS");
-              featuresCpC.add(cShape + "-PSEQcS");
-            }
-            if (flags.strictGoodCoNLL && ! flags.removeStrictGoodCoNLLDuplicates) {
-              featuresCpC.add(pShape + '-' + cShape + "-PSEQpcS"); // Duplicate (in goodCoNLL orig, see -TYPES below)
+              out.addConcat(pShape, "-PSEQpS");
+              out.addConcat(cShape, "-PSEQcS");
             }
+            if (flags.strictGoodCoNLL && ! flags.removeStrictGoodCoNLLDuplicates)
```
AngledLuffa: The diff is a lot easier to read if there are fewer {} changes mixed in with the actual code changes :)

src/edu/stanford/nlp/ie/NERFeatureFactory.java
```diff
@@ -923,237 +1021,233 @@ else if (flags.useTitle2) {
...
+        out.addConcat(c.get(CoreAnnotations.AbgeneAnnotation.class), "-ABGENE");
+        out.addConcat(p.get(CoreAnnotations.AbgeneAnnotation.class), "-PABGENE");
+        out.addConcat(n.get(CoreAnnotations.AbgeneAnnotation.class), "-NABGENE");
       }
 
       if (flags.useABSTRFreqDict) {
-        featuresC.add(c.get(CoreAnnotations.AbstrAnnotation.class) + "-ABSTRACT" + c.get(CoreAnnotations.FreqAnnotation.class) + "-FREQ" + c.getString(CoreAnnotations.PartOfSpeechAnnotation.class) + "-TAG");
-        featuresC.add(c.get(CoreAnnotations.AbstrAnnotation.class) + "-ABSTRACT" + c.get(CoreAnnotations.DictAnnotation.class) + "-DICT" + c.getString(CoreAnnotations.PartOfSpeechAnnotation.class) + "-TAG");
-        featuresC.add(c.get(CoreAnnotations.AbstrAnnotation.class) + "-ABSTRACT" + c.get(CoreAnnotations.DictAnnotation.class) + "-DICT" + c.get(CoreAnnotations.FreqAnnotation.class) + "-FREQ" + c.getString(CoreAnnotations.PartOfSpeechAnnotation.class) + "-TAG");
+        out.add(c.get(CoreAnnotations.AbstrAnnotation.class) + "-ABSTRACT" + c.get(CoreAnnotations.FreqAnnotation.class) + "-FREQ" + c.getString(CoreAnnotations.PartOfSpeechAnnotation.class) + "-TAG");
```
AngledLuffa: A little curious why no love for the 6 length features

src/edu/stanford/nlp/ie/NERFeatureFactory.java
```diff
@@ -397,68 +396,180 @@ public void init(SeqClassifierFlags flags) {
    */
   @Override
   public Collection<String> getCliqueFeatures(PaddedList<IN> cInfo, int loc, Clique clique) {
-    Collection<String> features = Generics.newHashSet();
+    Set<String> features = Generics.newHashSet(100);
+    FeatureCollector c = new FeatureCollector(features);
```
AngledLuffa: I wish there were some way to specify that FeatureCollector needs to have the suffix set at least once.  Perhaps the initial suffix could be part of the constructor?  Not sure if that would actually be cleaner.

AngledLuffa:
Mostly it looks good.  I haven't tested it for speed but I believe you that
it helps.

I did make a few comments on places where I think functionality has
changed.  Please review those and we can work together to integrate this.

Thanks!

src/edu/stanford/nlp/ie/NERFeatureFactory.java
kno10: Oops. Yes.

src/edu/stanford/nlp/ie/NERFeatureFactory.java
kno10: I'll add `if (col != null)` but I do not know whether it was intentional to have null-key features there.

src/edu/stanford/nlp/ie/NERFeatureFactory.java
kno10: Yes, these got worse over time as the branches diverged I think. It would be easier if there were a consistent code style everywhere; such as always using {}. In my other pull request, I spend some hours yesterday trying to remove as many of such changes as I could find.

src/edu/stanford/nlp/ie/NERFeatureFactory.java
kno10: Largely because there are fewer of them. But it would only be consistent. I had considered a varargs call, but that again would create an object.

src/edu/stanford/nlp/ie/NERFeatureFactory.java
kno10: The suffix is valid for a context; I'd even rather un-set it afterwards to avoid programming errors.
It would likely be nicer to move the setSuffix call from
```
featuresCpC(cInfo, loc, c.setSuffix("CpC"));
```
into the beginning of method `featuresCpC`, and also add a `c.setSuffix(null);` to the end to make it an invalid value. But there is at least one method which is called twice with a different suffix and slightly different parameters.

src/edu/stanford/nlp/ie/NERFeatureFactory.java
AngledLuffa: Are you thinking of adding 5 & 6, or are you satisfied as is?

src/edu/stanford/nlp/ie/NERFeatureFactory.java
kno10: I don't think this code is commonly used, and in fact I also do not think it will work reliably, because of the `makeGenericKeyCache`, which will only populated once, and anything added afterwards will likely not work. Maybe you can discuss whether this can be removed. In CoreLabel some support for this seems to be already commented out.

src/edu/stanford/nlp/ie/NERFeatureFactory.java
kno10: Depends also on what you would prefer. Adding 5 and 6 probably means about 20 lines of code more to maintain. If these features are not used in any of the standard models, it may not be worth it.

AngledLuffa:
I would still make it handle "null" without crashing just in case.  I don't
think there's any guarantee on non-null at this point in the code.  I have
no idea if any existing models use this feature (you may be right that it
is unused)

AngledLuffa:
Either way is fine with me

kno10:
I have pushed a new version.
Instead of having 8 variants of the add() call, I adopt a builder design pattern:
`out.build().append("fragment").dash().append("second").add();`
similar to the `StringBuilder` API of core Java. The `begin()` method is mostly useless - it asserts that the buffer is currently empty. This helps to detect missing calls to add() - which can unfortunately be forgotten in this API. `.dash()` is just a shorthand for `.append('-')` here, used around 130 times. All these calls (except for add) are tiny, and should be immediately inlined by Java Hotspot.
I have not benchmarked this version, and hope I did not miss any add() calls myself (search and replace mostly, so this should be fine).

AngledLuffa:
Maybe reset() would be better than build(), but I have no strong feelings
about this

kno10:
Go ahead and rename it. I initially used `begin()`. Then switched to build because of the builder pattern. But it would there better fit the last "verb". `reset()` is more accurate.

AngledLuffa:
Merged.  Did not bother to rename the method

AngledLuffa:
Thanks!

PR #615, opened 2018-01-26
Speed up file parsing.

kno10:
Two improvements to file parsing (e.g. for CoNLL):

1. reuse Matcher and StringBuilder to reduce garbage collection
2. parse the column mapping (map parameter) only once into annotation classes


manning:
Thanks! Merging this in.

PR #625, opened 2018-02-15
Add taxonomy abbreviations to lexer

kopperud:
Hello, I have a feature request:

I work with taxonomic literature, and often this kind of sentences show up:

"Microporella cf. ciliata is found in India."
"Micropora sp. is a fossil."
"Mrgaretta aff. tenuis is another animal."
"Wanganuri gen. nov. is an invented genus."

These are standard latin abbreviations (https://en.wikipedia.org/wiki/List_of_Latin_abbreviations), and I would like for them to be tokenized as units. This way I can have more precise sentence boundaries.

This contribution is public domain. I ran all tests, one skipped and none failed. If there is any problem with this implementation I would like to cooperate to find a compromise or better solution.

Best regards,
Bjørn Tore Kopperud


manning:
Thanks, basically happy to merge, but I think there is a fine point with "cf." which was already there and the question is the default treatment in the heuristic sentence breaking.

manning:
Actually also same thing for "gen.".  But I have added "sp.", "aff." and "spp." where you had them.

PR #628, opened 2018-02-20
Optimize Viterbi implementation

kno10:
Optimize the Viterbi code by three key means:

* Avoid redundant array accesses and computations by using much local variables (yes, theoretically a good compiler would recognize and avoid these redundancies, but as far as I can tell, hotspot usually doesn't, and local references for nested arrays help)
* Break into multiple methods, so they can be optimized independently by hotspot. Hotspot works great with a single hot loop in a method, but has limitations if you have >2 hot loops in the same method (because it might optimize the first loop before it has enough/any statistics on the later loops).
* Instead of computing division and modulo, use division and manually compute remainder, since a-b*c will usually be faster than a modulo.

All of this is very technical, and not pretty (well, breaking into multiple methods is not bad). But as this is a very central method, such optimizations may be worth it.

### DIFF
--- a/src/edu/stanford/nlp/sequences/ExactBestSequenceFinder.java
+++ b/src/edu/stanford/nlp/sequences/ExactBestSequenceFinder.java
@@ -70,26 +70,30 @@ public class ExactBestSequenceFinder implements BestSequenceFinder  {
     }
 
     // Do forward Viterbi algorithm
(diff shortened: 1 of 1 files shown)


manning:
Thank you for this work on optimizing CoreNLP hotspots! Running this now on JDK11, the actual speedup seems to be very small – maybe the compiler has gotten better in the meantime – but it has to be an improvement to have that one huge method broken up!

PR #646, opened 2018-03-13
avoid memory leak

pva-01:
in long using term if exception appears environment will not be cleaned


manning:
Thanks!  I'll merge this.

PR #651, opened 2018-03-22
Add IsNewlineAnnotation in Chinese segmentation output

davecart:
Fixing issue where, in v3.9.1 (and possibly 3.9), Chinese segmentation output can no longer be passed to sentence splitting (e.g. annotators = segment, ssplit, pos). Adding in the CoreAnnotations.IsNewlineAnnotation.class in the segmentation output.

### DIFF
--- a/src/edu/stanford/nlp/pipeline/ChineseSegmenterAnnotator.java
+++ b/src/edu/stanford/nlp/pipeline/ChineseSegmenterAnnotator.java
@@ -425,10 +425,14 @@ public class ChineseSegmenterAnnotator implements Annotator  {
     if (separatorPattern.matcher(tokenText).matches()) {
       // Map to CoreNLP newline token
       tokenText = AbstractTokenizer.NEWLINE_TOKEN;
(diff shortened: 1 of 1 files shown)


J38:
Hi, the way to use the Chinese segmenter now is to call the tokenizer with these settings:

```tokenize.language = zh
segment.model = edu/stanford/nlp/models/segmenter/chinese/ctb.gz
segment.sighanCorporaDict = edu/stanford/nlp/models/segmenter/chinese
segment.serDictionary = edu/stanford/nlp/models/segmenter/chinese/dict-chris6.ser.gz
segment.sighanPostProcessing = true```

These are the settings in the Chinese properties file.  A user should not directly use the segment annotator, as it is just a sub-annotator for the main tokenizing annotator.

Sorry for any confusion!

J38:
I think it could make sense if I changed those properties to tokenize.segment...
Also, a lot of annotators are starting to have sub-annotators, such as this example.  I am definitely hoping to add a more formal notion of "sub-annotator" to the code at some point...

PR #652, opened 2018-03-22
xception in thread "main" java.lang.IllegalArgumentException: No annotator named mention

ammarasmro:
I believe that there has been a change in the way the mention annotator is called. 
It isn't working anymore without prefixing it with coref.

If you know that this is going to be the standard way of calling the mention annotator please let me know so that I can find other mentions of that annotator and change them


J38:
Hi there have been some changes to naming conventions.  The `coref` annotator should run the mention annotator by default now.  And yes, the `mention` annotator has had it's name changed to `coref.mention` to make it clearer it is generating coref mentions.  Here is a link to the migration guide: https://stanfordnlp.github.io/CoreNLP/migration.html

PR #654, opened 2018-03-23
Amend abbreviation for subspecies

kopperud:
Species are abbreviated *sp.*
Subspecies are abbreviated *ssp.*, sometimes *subsp.* - but never *spp.*

https://en.wikipedia.org/wiki/Subspecies


manning:
Not my area of expertise, but isn "spp" used as a plural? That's what Wikipedia says!

    https://en.wikipedia.org/wiki/Binomial_nomenclature#Writing_binomial_names

But happy to add these other two as well.

PR #659, opened 2018-03-31
TokensRegexNERAnnotator | Implemented per-file header declarations and header=true 

alsora:
It is possible to specify a different header for each mapping file: it is handled as other fields by the processPerFileOptions method at line [885](https://github.com/alsora/CoreNLP/blob/master/src/edu/stanford/nlp/pipeline/TokensRegexNERAnnotator.java#L885).

Here is an example of mapping files with header declaration:

`properties.setProperty("ner.fine.regexner.mapping", 
"ignorecase=true, header=pattern overwrite priority, file1.tab;" +
"ignorecase=true, file2.tab");
`

The header items are whitespace separated and can also be specified for one of the files.

Files MUST be separated with semi column.

In the same way, it is possible to fetch the header from the first line of a specific file.

`properties.setProperty("ner.fine.regexner.mapping", 
"ignorecase=true, header=true, file1.tab;" +
"ignorecase=true, header=pattern overwrite priority group, file2.tab");
`

When reading entries from a file a check is present in order to discard the first line if it's the header (line [690](https://github.com/alsora/CoreNLP/blob/master/src/edu/stanford/nlp/pipeline/TokensRegexNERAnnotator.java#L690)).

Then it has been implemented the possibility of setting  "ner.fine.regexner.mapping.header=true" in line [250](https://github.com/alsora/CoreNLP/blob/master/src/edu/stanford/nlp/pipeline/TokensRegexNERAnnotator.java#L250).
This has the same effect of writing "header=true" for each provided mapping file.

### DIFF
--- a/src/edu/stanford/nlp/pipeline/TokensRegexNERAnnotator.java
+++ b/src/edu/stanford/nlp/pipeline/TokensRegexNERAnnotator.java
@@ -257,8 +257,8 @@ public class TokensRegexNERAnnotator implements Annotator  {
     				mappingLine = "header=true, "+ mappingLine;
     				mappings[i] = mappingLine;
     			}
-    			else if (!mappingLine.toLowerCase().matches("header\\s*=\\s*true")){
-    				throw new Error("Error! The annotator header property is set to true, but a different options has been provided for mapping file: " + mappingLine);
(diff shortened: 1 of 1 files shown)


manning:
Thanks. Will merge.

PR #703, opened 2018-05-14
Update corenlp dedicated server start script to automatically recover from system crash

zyshin:
Update start and restart scripts to handle isolated .shutdown file caused by unexpected service stop so as to allow automatic recover;  add status command.

### DIFF
--- a/src/edu/stanford/nlp/pipeline/demo/corenlp
+++ b/src/edu/stanford/nlp/pipeline/demo/corenlp
@@ -22,23 +22,37 @@
 #
 SERVER_USER="nlp"
 CORENLP_DIR="/opt/corenlp"
+PROP_FILE="$CORENLP_DIR/server.properties"
+LISTEN_PORT=9000
+JVM_MAX_MEMORY=8g
 
 
 do_start()
 {
   if [ -e "$CORENLP_DIR/corenlp.shutdown" ]; then
     echo "CoreNLP server is already running!"
-    echo "If you are sure this is a mistake, delete the file:"
-    echo "$CORENLP_DIR/corenlp.shutdown"
+    # echo "If you are sure this is a mistake, delete the file:"
+    # echo "$CORENLP_DIR/corenlp.shutdown"
+    echo "Force restarting"
+    do_restart
   else
     export CLASSPATH=""
     for JAR in `find "$CORENLP_DIR" -name "*.jar"`; do
       CLASSPATH="$CLASSPATH:$JAR"
     done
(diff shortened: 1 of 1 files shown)


src/edu/stanford/nlp/pipeline/demo/corenlp
```diff
@@ -22,23 +22,37 @@
 #
 SERVER_USER="nlp"
 CORENLP_DIR="/opt/corenlp"
+PROP_FILE="$CORENLP_DIR/server.properties"
+LISTEN_PORT=9000
+JVM_MAX_MEMORY=8g
```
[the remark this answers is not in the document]
zyshin: The only concern is that 8g is more affordable for users to try this demo on their own PCs.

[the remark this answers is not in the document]
mlucas-NU: It's not a bad idea, but I don't think it's related to the purpose of your PR (to recover from crashes). Perhaps it's better suited for its own PR or one related to memory management?

src/edu/stanford/nlp/pipeline/demo/corenlp
```diff
@@ -22,23 +22,37 @@
...
 
 
 do_start()
 {
   if [ -e "$CORENLP_DIR/corenlp.shutdown" ]; then
     echo "CoreNLP server is already running!"
-    echo "If you are sure this is a mistake, delete the file:"
-    echo "$CORENLP_DIR/corenlp.shutdown"
+    # echo "If you are sure this is a mistake, delete the file:"
```
sepe81: Better remove those two lines, instead of commenting out. (You're using Git.)

src/edu/stanford/nlp/pipeline/demo/corenlp
```diff
@@ -56,15 +70,21 @@ do_restart()
...
 start) do_start
 ;;
+status) do_status
+;;
 stop) do_stop
 ;;
 restart) do_restart
 ;;
 esac
-
```
sepe81: better restore the EOL (newline) here, see https://unix.stackexchange.com/a/18789

src/edu/stanford/nlp/pipeline/demo/corenlp
zyshin: fixed

src/edu/stanford/nlp/pipeline/demo/corenlp
zyshin: fixed

PR #709, opened 2018-05-28
Sentiment Pipeline: Disable builgraphs for parser

singh-abhijeet:
Optimization: Set `parse.buildgraphs` to `false` for not building Enhanced Deps (enhancedDeps, enhanced++Deps) as it is not utilized by Sentiment Pipeline internally.

**Benefits**

- Functionally optimized as extra Dependencies are not generated by the Parser

- Lightweight response body while using Sentiment Annotator via CoreNLPServer 

- Modular code change (only in SentimentPipeline.java) and thus Parser  behavior is not impacted for all Pipelines but Sentiment.


manning:
Thanks! Will merge.

PR #726, opened 2018-06-19
Fix expected/actual ordering in assert

joehalliwell:
This corrects the error message in case of failure.


manning:
Thanks, merging.

PR #727, opened 2018-06-19
Fix percent runs

joehalliwell:
This fixes an issue where consecutive PERCENT tokens e.g. "12% 13%" get collapsed into a single entity mention. There's a similar problem with MONEY which I hope to tackle in due course.


manning:
Thanks! Will merge.

PR #805, opened 2018-11-30
Add minimal root checks to prevent runtime exception

kphaison:
Other methods in the UniversalEnglishGrammaticalStructure class check if the SemanticGraph variable has roots before executing Semgrex Pattern matching and does nothing if there are no roots; the Semgrex Pattern requires that graphs have a root to operate. 
Current code for 5 methods are missing these guard statements, only 2 have been fixed as they are not proceeded by checks in other methods calling them.


manning:
Thanks!

PR #810, opened 2018-12-23
improve Optional usage in CoreQuote

apulbere:


manning:
Yes, that's much cleaner, thanks!

Issue #799, opened 2018-11-16
Using DepParse in conjunction with Quote in 3.9.x


aendrew:
I'm using poorna-kumar/gendermeme-core in conjunction with CoreNLP and since upgrading to the latest version I am unable to make anything work due to the change to the Quote annotator.

The entirety of the annotator loadout I need is

```
['pos', 'lemma', 'ner', 'parse', 'depparse', 'dcoref', 'quote', 'openie']
```

(Which given that the first few get pulled in as deps can be simplified to `['depparse', 'dcoref', 'coref', 'openie', 'quote']`) ...but it seems the Quote and DepParse annotators are wholly incompatible.

I suppose I can downgrade to 3.8.x but I'd rather not if possible... Any suggestions? Thanks!

J38:
Could you elaborate on what kind of error you are seeing...depparse should definitely work fine with quote.

J38:
For instance this pipeline should work fine:

```
tokenize,ssplit,pos,lemma,ner,depparse,coref,natlog,openie,quote
```

sillystring13:
I think that the Quote Annotator is forcing the default depparse model here:
https://github.com/stanfordnlp/CoreNLP/blob/eb43d5d9150de97f8061fa06b838f1d021586789/src/edu/stanford/nlp/quoteattribution/QuoteAttributionUtils.java#L223 

I can't see anywhere in the code that it's checking the depparse.model property, so if he's got a custom depparse in the pipeline it might conflict.

sillystring13:
I think that the properties for QuoteAttributionAnnotator are getting clobbered in the current flow.  Here's what I've traced through

AnnotationImplementor passes properties quote.* when it creates QuoteAnnotator
https://github.com/stanfordnlp/CoreNLP/blob/eb43d5d9150de97f8061fa06b838f1d021586789/src/edu/stanford/nlp/pipeline/AnnotatorImplementations.java#L264-L268

QuoteAnnotator creates the QuoteAttributionAnnotator using the truncated properties
https://github.com/stanfordnlp/CoreNLP/blob/eb43d5d9150de97f8061fa06b838f1d021586789/src/edu/stanford/nlp/pipeline/QuoteAnnotator.java#L171

[...]

PR #819, opened 2019-01-10
Fix quoteattribution properties

sillystring13:
Patch to fix problem with QuoteAttributionAnnotator not using properties correctly.  The default locations get used and any custom properties for quote.attribution.* and depparse.* are currently ignored.   

The fix
- passes through relevant properties (quote.*, depparse.*) to QuoteAnnotator so that it can configure QuoteAttributionAnnotator correctly
- updates doc to reflect property quoteattribution is actually quote.attribution (to fulfill passthrough on quote.*)
- updated QuoteAttributionUtil so that it uses the properties passed through for depparse.* and creates a DependencyParser based on those properties instead of default only. 

I tested this locally with the default properties and file locations, as well as a test that manually loaded the files from /tmp/edu/*  (ex. properties.put("quote.attribution.familyWordsFile", "/tmp/edu/stanford/nlp/quoteattribution/family_words.txt");

This pull request is related to the discussion in https://github.com/stanfordnlp/CoreNLP/issues/799 which I joined because it was similar to my problem.  OP problem ultimately had a different source, but I wrote this fix to solve my problem.   My original problem was that I couldn't use quote attribution if my model files weren't in the classpath because the paths in properties weren't getting used.  

I declare this contribution in the public domain.

### DIFF
--- a/src/edu/stanford/nlp/pipeline/AnnotatorImplementations.java
+++ b/src/edu/stanford/nlp/pipeline/AnnotatorImplementations.java
@@ -262,8 +262,14 @@ public class AnnotatorImplementations  {
    * Annotate quotes and extract them like sentences
    */
   public Annotator quote(Properties properties) {
-    Properties relevantProperties = PropertiesUtils.extractPrefixedProperties(properties,
-        Annotator.STANFORD_QUOTE + '.');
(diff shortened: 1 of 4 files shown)


J38:
Thank you for fixing these issues !

PR #842, opened 2019-03-01
Needed to follow redirects

Nepomuceno:
curls does not do it by default so adding the flag -L


manning:
Thanks, yes, we moved files to a new download machine.

PR #865, opened 2019-04-09
grammar

jsoref:
suggestions by app.grammarly.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Maven Central](https://img.shields.io/maven-central/v/edu.stanford.nlp/stanford-corenlp.svg)](https://mvnrepository.com/artifact/edu.stanford.nlp/stanford-corenlp)
 [![Twitter](https://img.shields.io/twitter/follow/stanfordnlp.svg?style=social&label=Follow)](https://twitter.com/stanfordnlp/)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,23 +4,23 @@
...
 
 The Stanford CoreNLP code is written in Java and licensed under the GNU General Public License (v3 or later). Note that this is the full GPL, which allows many free uses, but not its use in proprietary software that you distribute to others.
 
 ### Build Instructions
 
 Several times a year we distribute a new version of the software, which corresponds to a stable commit.
 
-During the time between releases, one can always use the latest, under development version of our code.
+During the time between releases, one can always use the latest, under development, version of our code.
```
AngledLuffa: This seems wrong

AngledLuffa:
Merged the two grammar changes I agreed with

README.md
jsoref: To be fair, the internet doesn't agree w/ my suggestion, it also doesn't like the current string either: https://www.google.com/search?q=%22latest%2C+under+development%22

The reason I proposed adding the comma is because w/o it, the parsing I encountered is:

> During the time between releases (one can always use the latest) under development version of our code.

... which doesn't make sense.

I think the correct approach is just:

> During the time between releases one can always use the latest under development version of our code.

PR #871, opened 2019-04-29
V3.9.2

phyank:
FIX: make special filename '-' compatible with option arguments.

According to the description on the official doc, filename "-" means that the parser would read from stdin and write to the stdout. However, in the previous implementation, if the first argument is a option argument start with "-", the main() will take all arguments start with "-" as option arguments, including the "-".

After this fix, the program can accept "-" as a filename with other  option arguments. I think it make little effect on the performance because this code fragment won't run frequently, and it won't parse thousands of option argument either.

I've recompiled the project and the new version of LexicalizedParser.java works fine.


phyank:
Why did the building process fail? I think it's not due to my change

AngledLuffa:
You ruined everything!

Nah, just kidding.  I merged this internally.

PR #872, opened 2019-05-01
Rejuvenate log levels

khatchad:
## Introduction

We are in the process of evaluating our [research prototype Eclipse plug-in](https://github.com/ponder-lab/Logging-Level-Evolution-Plugin) that "rejuvenates" log statement levels based on how "interesting" the enclosing methods are to the developers. We hypothesize that portions of a system that are worked on more and more recently should have higher log levels (e.g., INFO as compared to FINEST) and vice-versa. This places event logs related to tasks that are currently more important to the forefront while pushing event logs pertaining to tasks worked on in the past to the background. The goal is to reduce information overload, bring more relevant events to developers' attention, and alleviate developers from manually making log level changes.

The transformation decision is made by analyzing the "degree of interest" (DOI) values of enclosing methods for logging invocations. DOI value is a kind of real number for a program element which shows how developers are interested in it. It is computed from the interaction events between developer and program elements, such as edits. In this project, we calculate the DOI using the project's git history.

We are looking for feedback on our tool from developers. If you can, we would appreciate if you can **comment on each of the transformations** in the case that this PR is not accepted. Of course, we would also love to contribute to your project.

## Settings

We have several analysis settings. We can vary these settings and rerun if you desire. The settings we are using in this pull request are:

1. Treat CONFIG, WARNING, and SEVERE levels as a category and not a traditional level, i.e., our tool ignores these log levels.
1. Never lower the logging level of logging statements within catch blocks.
1. Never lower the logging level of logging statements within if statements.
1. Avoid log wrapping by disregarding logging statements contained in if statements mentioning log levels.
1. The greatest number of commits evaluated: 6000.

The head at time of analysis was: 5fdbfb209069276e95e1765093df9855d2cf2c38

### DIFF
--- a/src/edu/stanford/nlp/dcoref/CoNLL2011DocumentReader.java
+++ b/src/edu/stanford/nlp/dcoref/CoNLL2011DocumentReader.java
@@ -90,7 +90,7 @@ public class CoNLL2011DocumentReader  {
       Collections.sort(this.fileList);
     }
     curFileIndex = 0;
-    logger.info("Reading " + fileList.size() + " CoNll2011 files from " + filepath);
+    logger.finest("Reading " + fileList.size() + " CoNll2011 files from " + filepath);
   }
 
   private static List<File> getFiles(String filepath, Pattern filter)
@@ -121,7 +121,7 @@ public class CoNLL2011DocumentReader  {
         docIterator = new DocumentIterator(curFile.getAbsolutePath(), options);
       }
       while ( ! docIterator.hasNext()) {
-        logger.info("Processed " + docIterator.docCnt + " documents in " + curFile.getAbsolutePath());
+        logger.finest("Processed " + docIterator.docCnt + " documents in " + curFile.getAbsolutePath());
         docIterator.close();
         curFileIndex++;
         if (curFileIndex >= fileList.size()) {
@@ -131,7 +131,7 @@ public class CoNLL2011DocumentReader  {
         docIterator = new DocumentIterator(curFile.getAbsolutePath(), options);
       }
       Document next = docIterator.next();
-      SieveCoreferenceSystem.logger.fine("Reading document: " + next.getDocumentID());
+      SieveCoreferenceSystem.logger.finest("Reading document: " + next.getDocumentID());
       return next;
     } catch (IOException ex) {
       throw new RuntimeIOException(ex);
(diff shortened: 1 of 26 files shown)


khatchad:
Seems that the build error also exists in 5fdbfb209069276e95e1765093df9855d2cf2c38, which is the version analyzed.

src/edu/stanford/nlp/dcoref/CoNLL2011DocumentReader.java
```diff
@@ -971,7 +971,7 @@ public static void main(String[] args) throws IOException {
       System.exit(-1);
     }
     PrintWriter fout = new PrintWriter(outfile);
-    logger.info("Writing to " + outfile);
+    logger.finest("Writing to " + outfile);
```
kno10: This is a pretty good counter-example to your hypothesis...

khatchad: Thanks for the feedback @kno10. I believe our hypothesis may only apply to a certain class of logs and not all logs in general. Nevertheless, we are [working on making the transformations a bit less drastic](https://github.com/ponder-lab/Logging-Level-Evolution-Plugin/issues/232).

khatchad:
Thanks, @AngledLuffa. It's possible that the tool would also increase the log levels, but we found some bugs we are working out, as I mentioned [above](https://github.com/stanfordnlp/CoreNLP/pull/872#issuecomment-532782793). Thanks again for your feedback.

kno10:
On a related note - I've seen a number of `System.err.println`s pop up again, that may better be converted into logging statements. Usually guarded with debug flags though (this does however show that the custom "Redwood" logging system may not be up to standards - usually one would rather use the logging levels to configure the granularity instead of having to edit the source code).
Example (inserted just 8 months ago):
https://github.com/stanfordnlp/CoreNLP/blob/8da1176c52d37eb2d3a4a3b2323e90ce6cf69db7/src/edu/stanford/nlp/tagger/maxent/TestSentence.java#L391-L396
https://github.com/stanfordnlp/CoreNLP/blob/8da1176c52d37eb2d3a4a3b2323e90ce6cf69db7/src/edu/stanford/nlp/tagger/maxent/TestSentence.java#L584
A tool that automatically suggests which `System.err.println` to convert to a logging statement would likely be more useful. ;-)

khatchad:
Thanks for the info @kno10, we'll have a look at this.

PR #889, opened 2019-06-01
Fixed list index for 10th token

ikwattro:


manning:
Debatable depending on how you interpret "10th" in comment, but okay, I'll merge it.....

Issue #938, opened 2019-08-22
Nondeterminism in parse annotator results


alexey-tarasov-irl:
I'm running into a problem, where `parse` annotator (with `englishRNN.ser.gz` model) doesn't always produce the same dependency graph for the same input. For instance, if I annotate a sentence "X should not be 0." ten times in a loop, each iteration can end with one of the following parses (the difference is marked in **bold**):

-> be/VB (root)
  -> X/NNP (nsubj)
  -> should/MD (aux)
  -> not/RB (neg)
  **-> 0/CD (xcomp)**
  -> ./. (punct)

-> be/VB (root)
  -> X/NNP (nsubj)
  -> should/MD (aux)
  -> not/RB (neg)
  **-> 0/CD (nmod:tmod)**
  -> ./. (punct)

[Here is the code to reproduce the issue](https://github.com/stanfordnlp/CoreNLP/files/3529531/Determinism.txt).

I looked into it, and it seems like it might be caused by a rounding error. From what I see in debugger, `parse` method of `RerankingParserQuery` class first gets a list of parses and then reranks them. In `bestKParses` both annotations are present, but the one with 0-xcomp is always ranked above one with 0-nmod:tmod (parses in question are 0 and 15):

![image](https://user-images.githubusercontent.com/54396013/63502724-0d774600-c4c6-11e9-97dc-4b1c6f2d3faa.png) 

Then, these parses are getting re-ranked by the following line in `rerank` method of `RerankingParserQuery`:

`score = op.baseParserWeight * score + rerankerQuery.score(scoredTree.object());`

The first term is the score contained in `bestKParses`, but since `score` factor is equal to zero in `englishRNN.ser.gz` model, the new score is effectively the score given by `rerankerQuery`.  The two annotations I'm concerned about seem to be getting the same score, resulting in a tie. It's not a problem, because `Collections.sort()` is stable, so the correct annotation ends up at the top and is getting returned as the resulting annotation by `getBestParse` method (see 0 and 3):

![image](https://user-images.githubusercontent.com/54396013/63503385-449a2700-c4c7-11e9-96fc-d172683c6af3.png)

[...]

PR #946, opened 2019-09-17
Nondeterminism in parse annotator results #938

alexey-tarasov-irl:
Fixing a floating-point-related bug, which triggered non-determinism described in #938.


alexey-tarasov-irl:
> Do you think these changes need to be pushed further down into the score functions? There are lots of matrix operations using floating points in this module. Was the only issue you found the different ordering as a result of the Set?

Non-deterministic ordering was the only issue I encountered. For this particular issue, it seems to me that there is no need for going deeper, as the scores themselves were deterministic (i.e. the same in both cases). But you're right, issues like this might be hiding somewhere else in the codebase as well.

AngledLuffa:
Merged internally.

PR #956, opened 2019-10-21
Set default value of entityMentionsLanguage

jamesdbaker:
entityMentionsLanguage has no default at the moment when created via a no-args constructor, but is required to be non-null. This PR sets the default language to "en".


AngledLuffa:
How does this come up?

AngledLuffa:
Merged, since it seems reasonable regardless

jamesdbaker:
> How does this come up?

I'm working on integrating certain CoreNLP components into a different pipeline (i.e. not the CoreNLP pipeline), and was trying to create the component on it's own just using the default config.

Thanks for merging.

PR #965, opened 2019-11-13
Remove unused dependency

cesarsotovalero:
Hello. I noticed that dependency `org.apache.lucene:lucene-queries:7.5.0` is declared in the project's `pom` but it is not used. Hence, this dependency can be safely removed to make CoreNLP slimmer and its Maven dependency tree less complex.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,12 +68,6 @@
 
     <dependency>
       <groupId>org.apache.lucene</groupId>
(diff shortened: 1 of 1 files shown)


AngledLuffa:
Thank you for pointing this out.  There are some other adjustments we needed to make internally, but I have integrated this change.

Issue #991, opened 2020-02-04
StanfordCoreNLPClient fails "silently"


drewkerrigan:
In `StanfordCoreNLPClient.annotate`, if the server is not running, or some other exception occurs in this catch block: https://github.com/stanfordnlp/CoreNLP/blob/master/src/edu/stanford/nlp/pipeline/StanfordCoreNLPClient.java#L461-L467 - the exception is logged but the calling function never finds out about it. This is problematic because we end up with an annotation that wasn't annotated, but no way to test if it failed later on.

AngledLuffa:
I merged your change.  Thank you

PR #992, opened 2020-02-04
Add an ExceptionAnnotation to failed annotations in StanfordCoreNLPClient

drewkerrigan:
Should resolve https://github.com/stanfordnlp/CoreNLP/issues/991

This change allows a calling function to inspect and react to exceptions that occur when annotating using an external server. For example:

```
final Annotation annotation = new Annotation(text);

AnnotationPipeline pipeline = new StanfordCoreNLPClient(props, host, port, threads, key, secret);
pipeline.annotate(annotation);

if (annotation.containsKey(CoreAnnotations.ExceptionAnnotation.class)) {
  final Throwable t = annotation.get(CoreAnnotations.ExceptionAnnotation.class);
  throw new RuntimeException(t);
}
```

This contribution can be regarded as "de minimis" since it is less than 6 lines changed. This contribution is also in the public domain.

### DIFF
--- a/src/edu/stanford/nlp/pipeline/StanfordCoreNLPClient.java
+++ b/src/edu/stanford/nlp/pipeline/StanfordCoreNLPClient.java
@@ -3,6 +3,7 @@ package edu.stanford.nlp.pipeline;
 import edu.stanford.nlp.io.FileSequentialCollection;
 import edu.stanford.nlp.io.IOUtils;
 import edu.stanford.nlp.io.RuntimeIOException;
+import edu.stanford.nlp.ling.CoreAnnotations;
 import edu.stanford.nlp.util.StringUtils;
 import edu.stanford.nlp.util.logging.Redwood;
 import edu.stanford.nlp.util.logging.StanfordRedwoodConfiguration;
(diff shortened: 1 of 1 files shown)


src/edu/stanford/nlp/pipeline/StanfordCoreNLPClient.java
```diff
@@ -464,6 +465,8 @@ public void annotate(final Annotation annotation, final Consumer<Annotation> cal
           log.info("Trying to annotate locally...");
           StanfordCoreNLP corenlp = new StanfordCoreNLP(properties);
           corenlp.annotate(annotation);
+        } else {
+          annotation.set(CoreAnnotations.ExceptionAnnotation.class, t);
```
AngledLuffa: Looks reasonable, although it looks like the ExceptionAnnotation didn't make it in to the code change

AngledLuffa: Thanks!

drewkerrigan: @AngledLuffa how do you mean it didn't make it into the code change?

AngledLuffa:
Oh, my mistake.  I thought the ExceptionAnnotation was new.

Thanks for the addition!

PR #1008, opened 2020-03-03
Ignore eclipse files

asashour:
This adds Eclipse specific files to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 *.so
 .DS_Store
 .settings/**
+/target/
+.classpath
+.project


.gitignore
```diff
@@ -6,3 +6,6 @@
 *.so
 .DS_Store
 .settings/**
+/target/
+.classpath
```
AngledLuffa: Our approach has been to include our own suggested .classpath in the root directory.  That wouldn't work if we make this change

asashour: Well, the change ignores those files from being marked for commit.

So, anyone can use his/her own `.classpath` without affecting the others.

I understand this PR doesn't have an effect, because `.classpath` is not currently in the github repository.

AngledLuffa:
Ultimately the internal version of .gitignore can't have .classpath in it,
but I'm not sure what the consequences are of changing the public
.gitignore and not the internal one.  I'll leave it open but there's no
expectation of merging it any time soon.

asashour:
Thanks, but no point to leave the PR open. I can change the local `.gitignore` and ensure it is not committed.

The main point was having all the classes files in `target` marked as changed, but it can be handled.

PR #1009, opened 2020-03-03
Replace deprecation usage of IOUtils methods

asashour:
I am not sure about the deprecation policy, should I delete those two methods as well?

I declare that the contribution is in the public domain.


AngledLuffa:
Thank you; this has been merged.

PR #1010, opened 2020-03-03
Use StringBuilder instead of StringBuffer

asashour:
As you know, `StringBuilder` is the new way, as I don't think the synchronization feature of `StringBuffer` is needed.

I declare that the contribution is in the public domain.


src/edu/stanford/nlp/ling/tokensregex/parser/ParseException.java
```diff
@@ -89,7 +89,7 @@ private static String initialise(Token currentToken,
                            int[][] expectedTokenSequences,
                            String[] tokenImage) {
     String eol = System.getProperty("line.separator", "\n");
-    StringBuffer expected = new StringBuffer();
+    StringBuilder expected = new StringBuilder();
```
AngledLuffa: You're updating a bunch of autogen code...  it will just get undone the next time these tools get updated.  I'm pretty averse to separating this out myself

asashour: Oh, any hint about those tools, how can one run/change them?

AngledLuffa:
For example, the precompiled code is in

edu/stanford/nlp/trees/tregex/TregexParser.jj

Generated using javacc.  We tried to be nice and leave sample command lines:

// javacc
"-OUTPUT_DIRECTORY=projects/core/src/edu/stanford/nlp/trees/tregex"
projects/core/src/edu/stanford/nlp/trees/tregex/TregexParser.jj

guess that sort of gives away our internal paths, but whatever.

Here is an exciting discovery, btw.  The last time someone did an automated
sed replace or whatever it is you've done, they missed that they were
updating some autogen files.  So, for example, the original TregexParser.jj
still looked for edu.stanford.nlp.util.Function, even though that
functionality (ha, ha) was eventually incorporated into java.

Anyway, javacc supposedly has an option to use StringBuilder instead of
StringBuffer, but turning that option on has no effect on the
ParseException or TokenMgrError classes.  I filed an issue on the javacc
github.  Let's see what happens

https://github.com/javacc/javacc/issues/165

AngledLuffa:
We integrated the updated javacc, so if you're up for revisiting this old PR, we'd be able to integrate it now

AngledLuffa:
I rebased it against dev and merged it here:

https://github.com/stanfordnlp/CoreNLP/pull/1115

AngledLuffa:
Thanks!

PR #1011, opened 2020-03-03
Migrate tests to JUnit 4

asashour:
This the first part to migrate tests to JUnit 4 way, please advise if I should continue migrating all tests.

Afterwards, a second PR would follow to migrate to JUnit 5.

### DIFF
--- a/itest/src/edu/stanford/nlp/coref/hybrid/ChineseCorefBenchmarkSlowITest.java
+++ b/itest/src/edu/stanford/nlp/coref/hybrid/ChineseCorefBenchmarkSlowITest.java
@@ -9,15 +9,15 @@ import java.util.Properties;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import edu.stanford.nlp.util.BenchmarkingHelper;
-import junit.framework.TestCase;
+import org.junit.Test;
+
+import edu.stanford.nlp.coref.CorefProperties;
 import edu.stanford.nlp.io.IOUtils;
 import edu.stanford.nlp.io.RuntimeIOException;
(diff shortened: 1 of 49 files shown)


itest/src/edu/stanford/nlp/coref/hybrid/ChineseCorefBenchmarkSlowITest.java
```diff
@@ -1,24 +1,20 @@
 package edu.stanford.nlp.coref.hybrid;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.IOException;
-import java.io.PrintStream;
-import java.io.StringReader;
+import java.io.*;
```
AngledLuffa: Please don't do this.  I think I'm allergic to import *

asashour: Honestly, I am with what you say. But i found `.*` in various places in the code, so I changed the default Eclipse settings, since I don't want to be proposing so many changes.

I am reverting this, so all imports are individual.

AngledLuffa:
Conceptually this seems fine.  Let me take a little time to review the
actual changes

AngledLuffa:
Please squash, if you don't mind

asashour:
Squashed, migrated more tests, notable change in double assertion in `NERBenchmarkTestCase`, it was comparing `Object` with `Object`, now comparing `double` with `double` with 0 `delta`.

AngledLuffa:
I merged most of this into our codebase.  There were a couple issues, like some tests which aren't publicly released but include the tests you were changing.  Also it's worth noting that there's test/src which hasn't been updated to junit 4 at all

PR #1012, opened 2020-03-03
The static fields and methods should be accessed in a static way

asashour:
There were detected as Eclipse warnings.


AngledLuffa:
This isn't really a priority for us.  Our code has thousands of warnings
anyway...  sorting through the specific list someone else provides is
making work rather than making our lives better.  Maybe you could turn off
the warnings?

PR #1013, opened 2020-03-03
Fix ArrayList is a raw type

asashour:


AngledLuffa:
This has been merged.

I'm going to request, though, that we mostly stick to functional changes.  I don't want to start rejecting changes but this is kind of make-work for us

PR #1014, opened 2020-03-04
Use .entrySet() instead of keySet()

asashour:
This way, it is faster, since map is searched for an entry only not, and not twice (one for key and another for value).

This PR is initially for discussing the change, then it would be followed by other commits/PRs to cover the whole code.

P.S.: I wonder if making `edu.stanford.nlp.stats.Counter` a `Map` would add a value.

### DIFF
--- a/itest/src/edu/stanford/nlp/dcoref/DcorefExactOutputITest.java
+++ b/itest/src/edu/stanford/nlp/dcoref/DcorefExactOutputITest.java
@@ -4,10 +4,11 @@ import java.io.BufferedWriter;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Properties;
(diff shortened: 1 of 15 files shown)


itest/src/edu/stanford/nlp/dcoref/DcorefExactOutputITest.java
```diff
@@ -4,10 +4,11 @@
 import java.io.FileWriter;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
```
AngledLuffa: Please keep functional changes (the Map.Entry change, for example) separate from syntactic sugar changes, which as I've mentioned elsewhere, we really don't want to spend effort on

AngledLuffa:
Using the entry set sounds useful - although obviously set lookups are not the expensive part of the algorithms in use here - but having several pieces of change mixed together like this makes it really unpleasant to have to sort through.  Please feel free to resubmit with just the functional changes.

PR #1015, opened 2020-03-04
Use .entrySet() instead of keySet()

asashour:


src/edu/stanford/nlp/loglinear/model/ConcatVectorNamespace.java
```diff
@@ -4,6 +4,7 @@
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Map.Entry;
```
AngledLuffa: I don't like this.  Map.Entry in the code is fine

itest/src/edu/stanford/nlp/dcoref/DcorefExactOutputITest.java
```diff
@@ -106,11 +106,10 @@ private static void saveResults(String filename, Map<Integer, CorefChain> chains
     FileWriter fout = new FileWriter(filename);
     BufferedWriter bout = new BufferedWriter(fout);
 
-    List<Integer> keys = new ArrayList<>(chains.keySet());
-    Collections.sort(keys);
+    Map<Integer, CorefChain> sorted = new TreeMap<>(chains);
```
AngledLuffa: I don't expect this to be as optimized as to Collections.sort.  Seems like an unnecessary change

asashour: It's a test, anyhow reverted.

src/edu/stanford/nlp/optimization/GoldenSectionLineSearch.java
```diff
@@ -161,10 +161,8 @@ public double minimize(DoubleUnaryOperator function) {
    * dump the {@code <x,y>} pairs it computed found
    */
   public void dumpMemory() {
-    Double[] keys = memory.keySet().toArray(new Double[memory.keySet().size()]);
-    Arrays.sort(keys);
-    for (Double key : keys) {
-      log.info(key + "\t" + memory.get(key));
+    for (Map.Entry<Double, Double> entry : new TreeMap<>(memory).entrySet()) {
```
AngledLuffa: Here is another use of sorting via TreeMap.  It's just a debug method, but I don't like the functionality change

AngledLuffa:
Alright, I merged this, but in general let me encourage you to spend your time on changes that make a more significant impact on the project

PR #1016, opened 2020-03-04
Add GeneratedCodeTest

asashour:
This ensures that no inadvertent changes go unnoticed.

The logic checks specific generated code against the timestamp of their source, however if a manual change was done later for only the generated one (e.g. by updating to CC library), then the `timestamp` must be explicitly put in the test.  The failed test gives a hint of what exactly should be done in this case.


AngledLuffa:
Although such a test would be useful, I don't like the idea of adding
eclipse packages.  Frankly it may simply be easier and just as helpful to
make sure all the files say DO NOT EDIT in big text across the top.

PR #1017, opened 2020-03-04
Fix (Penn)TreePrintTest on Windows

asashour:


AngledLuffa:
Thank you, this is quite helpful.  

In my defense, I've been doing almost all my development on linux lately.

PR #1020, opened 2020-03-07
Rectified the table formatting

dheerajmpai:
The table was not renderring properly on github. Have removed the additional unnecessary `|`. Now it looks proper.


AngledLuffa:
Thanks!

PR #1025, opened 2020-03-21
NERCombinerAnnotator: fix IndexOutOfBoundsException for empty text

aggarcia3:
For annotated empty text, the token list is empty. Therefore, assuming that there is at least one token in the list leads to exceptions. This assumption is made in the recently added transferNERAnnotationsToAnnotation method of NERCombinerAnnotator.

These changes check whether the token list is empty, and if it is then transferNERAnnotationsToAnnotation does nothing.


AngledLuffa:
Thanks for catching that.  I reversed the condition and checked that in, giving you credit (want to double check that it still solves the problem you originally found?)

aggarcia3:
Your changes still fix the problem for me. Thank you!

Issue #1022, opened 2020-03-19
Integration of Mocking framework


jvdassen:
Hello everyone, first off thank you for this project!

We are a group of students aiming at contributing more unit tests to this project. Thus we have read and acknowledged the [contribution rules](https://github.com/stanfordnlp/CoreNLP/blob/master/CONTRIBUTING.md).
During our analysis of the project we found different units that are hard to test since they depend on complex input objects, such as for example Files.

Would the maintainers of this project accept a pull request where the [Mockito library](https://site.mockito.org/) is integrated into the testing procedures to cover things that are hard to test and have 0% test coverage at the moment?
Looking at the licensing requirements of this project and the license of Mockito (MIT license) we assume that it would be compatible.

AngledLuffa:
In general we try to avoid expanding the library footprint of corenlp.  Do you have a few suggestions of classes you want tested more thoroughly?

PR #1031, opened 2020-04-21
Increase test coverage

jvdassen:
Hello everyone,

We have increase the scope of the unit testing in this project with focus on the `classify`, `math`, `process`, `simple` and `util` packages.
As you can see there were no modifications to the `src` of the project and we did not incorporate new dependencies as requested in [PR 1022](https://github.com/stanfordnlp/CoreNLP/issues/1022). 
All tests were written using the existing JUnit version and aligned to the formatting of the existing classes.

As requested in the [contribution guidelines](https://github.com/stanfordnlp/CoreNLP/blob/master/CONTRIBUTING.md) we are happy to declare this contribution under public domain.


AngledLuffa:
I'm going to start going through this, but I'd appreciate it if you could squash some of the changes down and set your whitespace to use 2 spaces instead of tabs.

test/src/edu/stanford/nlp/simple/TokenTest.java
```diff
@@ -79,13 +78,19 @@ public void testCharacterOffsetEnd() {
     }
 
     @Test
-    public void testWhiteSpace() {
+    public void testWhiteSpaceBefore() {
```
AngledLuffa, line 81: This really doesn't seem necessary

test/src/edu/stanford/nlp/dcoref/RulesTest.java
```diff
@@ -20,12 +19,12 @@
   List<CoreLabel> MIBM = SentenceUtils.toCoreLabelList("MIBM");
 
   public void testIsAcronym() {
-    assertTrue(Rules.isAcronym(IBM, IBM2));
-    assertTrue(Rules.isAcronym(IBM2, IBM));
-    assertFalse(Rules.isAcronym(IBM, IBMM));
-    assertFalse(Rules.isAcronym(IBM2, IBMM));
-    assertFalse(Rules.isAcronym(IBM, MIBM));
-    assertFalse(Rules.isAcronym(IBM2, MIBM));
+    assertTrue("Acronym IMB -> International Business Machines",Rules.isAcronym(IBM, IBM2));
```
AngledLuffa: You get a line number for where the test fails so I don't really agree this is an mystery assertion.  If you want to gather up a bunch of these changes into a single change, though, I can merge it

jvdassen:
Thanks for the feedback! Unfortunately well have to create a new PR to change the upstream branch to the one where we fixed and squashed our contributions!

PR #1033, opened 2020-04-21
Fix travis build (mvn install)

jvdassen:
It seems that the [master](https://github.com/stanfordnlp/CoreNLP/commits/master) is currently not successfully built on travis.
I think the version of the project was updated in `pom.xml` and the jar was deployed, however `travis.yml` still uses the old v3.9.2 causing it to fail during `mvn install`.


AngledLuffa:
Thank you, merged

PR #1036, opened 2020-04-24
Test coverage improvements

jvdassen:
Thanks for reviewing our [previous PR](https://github.com/stanfordnlp/CoreNLP/pull/1031). We have now aligned the formatting, reverted the tests you did not want and squashed all commits into a small number of commits.


test/src/edu/stanford/nlp/util/AcronymMatcherTest.java
```diff
@@ -19,6 +20,14 @@ public void testBasic() {
     assertTrue(AcronymMatcher.isAcronym("ECC", "Election Complaints Commission".split("\\s+")));
   }
 
+  @Test
```
AngledLuffa, line 23: Mind adding some comments for what this is testing?

test/src/edu/stanford/nlp/process/StopListTest.java
```diff
@@ -0,0 +1,33 @@
...
+	@Test
+	public void testGenericWordConstructor() {
+		// Tests that no exception being thrown during instantiation
+		StopList sut = new StopList();
+	}
+	@Test
+	public void testGenericWordConstainsString() {
+		// Tests that a common stopword is contained in the list as string
+		StopList sut = new StopList();
+		boolean containsWord = sut.contains("and");
```
AngledLuffa: What's the reason this isn't just one line, assertTrue(sut.contains("and"))

src/edu/stanford/nlp/patterns/TextAnnotationPatterns.java
```diff
@@ -41,10 +41,6 @@
   private static final Redwood.RedwoodChannels logger = Redwood.channels(TextAnnotationPatterns.class);
 
   public TextAnnotationPatterns() throws IOException {
```
AngledLuffa, line 43: I think the commented code is here so that someone working on this has a convenient suggestion on how to log intermediate results, but I could be wrong about this

test/src/edu/stanford/nlp/process/StemmerTest.java
```diff
@@ -0,0 +1,188 @@
+package edu.stanford.nlp.process;
```
AngledLuffa, line 1: To possibly save myself the hassle of checking myself, let me ask - have you compared these results to the results from a reference implementation such as the ones provided here?  https://tartarus.org/martin/PorterStemmer/

test/src/edu/stanford/nlp/process/WhitespaceTokenizerTest.java
```diff
@@ -33,6 +33,21 @@
     { "皇后", "後世", "and", "(800)\u00A0326-1456" },
   };
 
+  public static final String[] TEST_WHITESPACE_ONLY = {
```
AngledLuffa: I already merged this change, I believe

test/src/edu/stanford/nlp/process/WhitespaceTokenizerTest.java
```diff
@@ -61,5 +76,37 @@ public void testCoreLabelTokenizer() {
               (factory, true),
             TEST, RESULTS_EOL);
   }
+  
+  /* Tests the "whitespace only" bound */
```
AngledLuffa, line 80: oh, I see what happened.  I had renamed the test "testNoTextWhiteSpaceTokenizer"

test/src/edu/stanford/nlp/util/ArrayMapTest.java
```diff
@@ -30,10 +30,11 @@ public void setUp() {
     hmap.put("Baz", 500);
   }
  
-  public void testEquals() {
-    assertEquals(map, map);
+  public void testArrayMapEqualsMethod() {
```
AngledLuffa, line 33: testArrayMapEqualsHashMap probably sounds better.  I'll make that change myself

test/src/edu/stanford/nlp/objectbank/LineIteratorTest.java
```diff
@@ -4,32 +4,44 @@
 import java.util.Iterator;
 
 import org.junit.Assert;
+import org.junit.Before;
```
AngledLuffa, line 7: I wasn't sure what is being improved here - git commit comments would be a good place to explain that!  I merged the rest of the changes in this commit, though.

test/src/edu/stanford/nlp/process/WhitespaceTokenizerTest.java
```diff
@@ -98,17 +98,6 @@ public void testPureWhiteSpaceEOLTokenizer() {
     }
   }
 
-  /* Tests the "no whitespace at all" bound */
```
AngledLuffa: I see why the WhitespaceTokenizerTest showed up earlier.  Definitely could be squashed into the previous changes

test/src/edu/stanford/nlp/dcoref/RulesTest.java
```diff
@@ -20,12 +19,12 @@
   List<CoreLabel> MIBM = SentenceUtils.toCoreLabelList("MIBM");
 
   public void testIsAcronym() {
-    assertTrue(Rules.isAcronym(IBM, IBM2));
-    assertTrue(Rules.isAcronym(IBM2, IBM));
-    assertFalse(Rules.isAcronym(IBM, IBMM));
-    assertFalse(Rules.isAcronym(IBM2, IBMM));
-    assertFalse(Rules.isAcronym(IBM, MIBM));
-    assertFalse(Rules.isAcronym(IBM2, MIBM));
+    assertTrue("Acronym IMB -> International Business Machines", Rules.isAcronym(IBM, IBM2));
```
AngledLuffa, line 22: I like these comments a lot.  Thank you!

test/src/edu/stanford/nlp/ling/WordLemmaTagTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Test
+  public void testSetFromStringWordAndLemmaAndTag() {
+    WordLemmaTag wLT = new WordLemmaTag();
+    wLT.setFromString("studying/study/s");
+
+    checkWordLemmaTag(wLT, "studying", "study", "s");
+  }
+
+///// Failing TEST
```
AngledLuffa, line 61: Failing, in the sense that you expect it to have null values when set this way?

AngledLuffa:
I merged a bunch of these changes, and I left some comments on others.  At this point I think it would be helpful to once again ask you to rebase off the new master branch.  Thanks for the input!

test/src/edu/stanford/nlp/ling/WordLemmaTagTest.java
RinorSefa: The constructor of class WordLemaTag supports setting only the word value`  public WordLemmaTag(Label word) {
    this(word.value());
  }
` 
Also, the method setFromString is designed to support setting only the word variable as it can be noticed from the else branch. The bug with the method is that it never goes to the else branch because it passes the first if statement, and crashes at 4th line with StringIndexOutOfBoundsException Exception. This occurs because there are no dividers when we only set a word value. So the first and second variables represent the same value. 

`public void setFromString(String labelStr, String divider) {
    int first = labelStr.indexOf(divider);
    int second = labelStr.lastIndexOf(divider);
    if (first == second) {
      setWord(labelStr.substring(0, first));
      setTag(labelStr.substring(first + 1));
      setLemma(Morphology.lemmaStatic(labelStr.substring(0, first), labelStr.substring(first + 1)));
    } else if (first >= 0) {
      setWord(labelStr.substring(0, first));
      setLemma(labelStr.substring(first + 1, second));
      setTag(labelStr.substring(second + 1));
    } else {
      setWord(labelStr);
      setLemma(null);
      setTag(null);
    }
  }`

test/src/edu/stanford/nlp/process/StopListTest.java
jvdassen: For readability, but I agree it seems like overkill here.
We will replace with an assertion message and condense to one line which should still yield under 80 characters.

jvdassen: For readability, but I agree it seems like overkill here.
I replaced it with an assertion message and condensed to one line which should still yield under 80 characters.

test/src/edu/stanford/nlp/process/StemmerTest.java
jvdassen: I updated this: Now I test it against the official implementation of Martin Porters algorithm by himself, as done in commit [f35090a](https://github.com/stanfordnlp/CoreNLP/pull/1036/commits/f35090ae990699a477ad967c05a34aa98d560e2f).

test/src/edu/stanford/nlp/util/AcronymMatcherTest.java
AngledLuffa: A is a stopword, which is why the commented out line isn't working, right?

RinorSefa: yes, sorry my bad, will remove it!

test/src/edu/stanford/nlp/process/StopListTest.java
AngledLuffa: In that case, do you mind squashing that on your end?  It would make it easier to cherrypick one change at a time.

test/src/edu/stanford/nlp/util/AcronymMatcherTest.java
AngledLuffa: That's fine.  In fact I just left that test commented out with a note of why it wasn't working

AngledLuffa: Rebasing off the current master would make it easier to make more progress on this change.  Thanks for the contribution!

test/src/edu/stanford/nlp/process/StopListTest.java
jvdassen: Done! We squashed all these latest changes fixes together into one commit! Also we rebased it from your master.
Please tell us if you want them squashed into even larger/fewer commtis.

test/src/edu/stanford/nlp/util/AcronymMatcherTest.java
```diff
@@ -20,15 +20,6 @@ public void testBasic() {
     assertTrue(AcronymMatcher.isAcronym("ECC", "Election Complaints Commission".split("\\s+")));
   }
 
-  @Test
```
AngledLuffa, line 23: I think this change can be dropped, right?  It was a test added by your test

test/src/edu/stanford/nlp/objectbank/LineIteratorTest.java
```diff
@@ -4,32 +4,44 @@
...
-      Assert.fail("Probably too few things in iterator: " + e);
+                Assert.assertEquals("Wrong line", out, ans);
+            }
+        } catch (Exception e) {
+            Assert.fail("Probably too few things in iterator: " + e);
+        }
     }
-  }
 
+    @Test
```
AngledLuffa: This is changing the semantics of the test, I believe.  The final `if di.hasNext()` in the original version is checking that the iterator stops when there is no more input.  I don't see that happening here, though.

test/src/edu/stanford/nlp/simple/TokenTest.java
```diff
@@ -3,6 +3,8 @@
 import org.junit.Test;
 import static org.junit.Assert.*;
 
+import java.util.Optional;
```
AngledLuffa, line 7: Not seeing the reason for this being here - it compiles without this change, after all.  Was it intended to combine with the larger change at the end?

RinorSefa:
Yes, the change was dropped. It was a test method developed by us. Thank you for feedback
________________________________

test/src/edu/stanford/nlp/simple/SentenceTest.java
```diff
@@ -157,4 +167,120 @@ public void testTokenizeWhitespaceWithSpaces() {
...
+    /* The serialize method should write but should not close the
+     * provided OutputStream as specified in Sentence.java. */
+    boolean wasWrittenTo = mockedOutStream.getWriteCount() > 0;
+    boolean wasNotClosed = !mockedOutStream.gotClosed();
+
+    assertTrue(wasWrittenTo);
+    assertTrue(wasNotClosed);
+  }
+
+  @Test
```
AngledLuffa, line 234: This seems painful to change if the serialization changes.  Perhaps there's some alternate way of writing it where the object is written to a stream, deserialized, and compared with the original object

test/src/edu/stanford/nlp/simple/SentenceTest.java
```diff
@@ -157,4 +167,120 @@ public void testTokenizeWhitespaceWithSpaces() {
...
+  public void testDeserializeFromInputStream () throws IOException {
```
AngledLuffa, line 235: I really thought I just commented on this, but the comment is gone now - this seems difficult to change if the serialization format changes in the future.  I think it would work better if you first serialized a sentence, deserialized it, and compared the results.

Aside from that, I merged the changes for SentenceTest.  Thanks for the contribution!

test/src/edu/stanford/nlp/simple/TokenTest.java
```diff
@@ -112,4 +113,25 @@ public void testEndPosition() {
 
     assertEquals(35, token.endPosition());
   }
+
```
AngledLuffa, line 116: While this is technically correct, I think it would be more useful to put it in SentenceTest and check the after() of various values between 0 and n-1 as well

test/src/edu/stanford/nlp/util/AcronymMatcherTest.java
AngledLuffa: What I meant is, I merged that test.  Do you think it's better to remove it?  We can, if you like

test/src/edu/stanford/nlp/math/ADMathTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+  @Test
+  public void testMult() {
+    DoubleAD v1 = new DoubleAD(2.0, 4.0);
+    DoubleAD v2 = new DoubleAD(3.0, 2.0);
+
+    DoubleAD result = ADMath.mult(v1, v2);
+
+    DoubleAD correctResult = new DoubleAD(6.0, 16.0);
+    testCompareValues(correctResult, result, 0.0);
```
AngledLuffa, line 22: There's nothing about this class that makes it use exact floating point calculations, so I believe 0.0 is not appropriate for the allowed delta.

test/src/edu/stanford/nlp/ling/WordLemmaTagTest.java
AngledLuffa: Ideally this should be fixed then, right?

AngledLuffa: Also, it should be pointed out that commenting out the code isn't necessary... commenting out @Test already disabled the test.

AngledLuffa: Alright, I fixed the underlying bug and restored the test

AngledLuffa: There's a similar bug in the Factory.  Presumably the FactoryTest should be changed to expect null, null, right?  I'll check it in that way, and please confirm that's what you would expect.

AngledLuffa:
Alright, I merged in a bunch of these changes.  Your turn again, I think!  At this point it should be squashable to just one change if you rebase off master

test/src/edu/stanford/nlp/ling/WordLemmaTagTest.java
RinorSefa: Yes, there is a similar bug also in Factory. Thank you for solving it.

AngledLuffa:
Sounds good.  I just made that change.

AngledLuffa:
I think we are once again at a place where rebasing and squashing everything into one change would make it simpler to digest the remaining changes.  For example, a couple of the outstanding changes are a new file, followed by quite a few whitespace changes in the file.  Also, the first change of this PR is no longer needed.  

Thanks for the contribution!  Always good to increase test coverage, and it is nice to have fixed at least one bug because of the added tests.

jvdassen:
Thank you for your looking into this and for your feedback.
- We have squashed all commits ahead of CoreNLP/master into one remaining commit.
- We rebased ontop of CoreNLP/master.
- Regarding the `SentenceTest.java `, I totally agree that it violates the open/closed principle thus breaking the test once the protobuf serialization changes. I now use the strategy where I serialize and deserialize.

jvdassen:
@AngledLuffa As requesed, I removed the unused import statement as well as the duplicate test.

test/src/edu/stanford/nlp/math/ADMathTest.java
RinorSefa: Thank you for the feedback the code. ` assertEquals("Incorrect dot", correctResult.getdot(), result.getdot(), delta); ` will be changed to ` correctResult.equals(result) ` and `correctResult.equals(result,TOL)`  following the equals implemented in the DoubleAD class. The change will occur in the next rebase. Thank you for the review, and sorry for late reply.

jvdassen:
@AngledLuffa I also squashed into the same commit two additional changes:
- Testing the Ports class and fixing a small ambiguity
- Testing a large number of methods in the StringUtils class.

test/src/edu/stanford/nlp/util/IntTupleTest.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+  
+  @Test
+  public void testGetIntTuple() {
+    List<Integer> integers = new ArrayList<Integer>();
+    integers.add(0);
+    integers.add(1);
+    integers.add(3);
+    IntTuple tuple = IntTuple.getIntTuple(integers);
+    assertEquals(3, tuple.length());
```
AngledLuffa, line 78: Maybe throw in some tests that the values are expected as well?

test/src/edu/stanford/nlp/util/IntTupleTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @Before
+  public void setup() {
+    int[] nrs = {1, 2, 3, 5, 7, 11};
+    this.sut6Tuple = new IntTuple(nrs);
+    this.sut1Tuple = new IntTuple(13);
+  }
+
+  @Test
+  public void testCompareTo() {
```
AngledLuffa, line 24: Maybe a few comparisons of tuples of the same size would be of use as well.  Checks on tuples of different lengths where the initial item lists are the same and the sort is based on the length of the tuple.  Etc etc

jvdassen: Thanks @AngledLuffa I have increased the coverage of that case so that more cases/branches are tested.

AngledLuffa:
I was pretty sure I left a few other comments which have disappeared.  Did you see those other messages?

test/src/edu/stanford/nlp/simple/SentenceTest.java
AngledLuffa: One of the other comments I had made was that a ByteArrayOutputStream or some similar built in class should be sufficient for this.

jvdassen:
> I was pretty sure I left a few other comments which have disappeared. Did you see those other messages?

Yes! Sorry I didn't push all the changes.. But now the extension of the IntTupletest and the rework of the Sentence test using generic ByteArrayInput/Output was squashed into the two respective commits.
I know you left many more comments which somehow are not shown in this PR, but I have them and will ping the other contributors of this PR!

test/src/edu/stanford/nlp/util/AcronymMatcherTest.java
```diff
@@ -27,6 +26,13 @@ public void testFilterStopWords() {
     assertTrue(AcronymMatcher.isAcronym("AAAS", "American Association for the Advancement of Science".split("\\s+")));
   }
 
+  @Test
+  public void testIsAcronymWithBoundaries() {
+    assertFalse(AcronymMatcher.isAcronym("B", "".split("\\s+")));
```
AngledLuffa, line 31: Did you know that "".split("\\s+") is a list of length 1 with an empty string in it?  That was super confusing to debug

RinorSefa: @AngledLuffa. Yes I saw it. I just did not know if we should handle that, if yes where should we handle it?

AngledLuffa: ...and it would appear that although it is undocumented, the treatment of the empty string is to be a wildcard

AngledLuffa: Probably by passing in an empty list, since that seems to be the intent of the test

RinorSefa: Dear @AngledLuffa how would you like to continue? Should I change/drop the test?

AngledLuffa: As mentioned, an empty list would probably work the way you intended for this test.  Up to you whether that's worthwhile

PR #1046, opened 2020-05-27
Test coverage improvement using Mockito

jvdassen:
In issue #1022 we asked you about your willingness to integrate a mocking framework into the project.
This PR shows exemplary usage of the Mockito framework and how it can be used to test units that are hard to test otherwise.

As you can see everything has also been squashed down already into one commit.


AngledLuffa:
I would say yes, this looks interesting and worthwhile.  If you don't mind,
please add some documentation on what is being tested and how.  For
example, I'm not sure what is going on in DocumentTest.  Thanks!

PR #1051, opened 2020-06-03
Use ClassLoader.getSystemResourceAsStream(name) for loading models

bamthomas:
instead of     
```
IOUtils.class.getClassLoader().getResourceAsStream(name);
```

As the NLP pipelines are diverse and provide different results, we made them loadable as plugins. So we are using dynamic class loading using a [DynamicClassLoader](https://stackoverflow.com/a/59743937/396155) and setting the property `java.system.class.loader` to it.

The problem is that we have now a hierarchy of 3 class loaders : 
```
sun.misc.Launcher$ExtClassLoader
                       |
sun.misc.Launcher$AppClassLoader
                       |
org.icij.datashare.DynamicClassLoader
```
`IOUtils.class.getClassLoader()` is returning `AppClassLoader` but the models are visible in the ̀`DynamicClassLoader` via the class path of the application.

ClassLoaders can see the parents resources but cannot see the children resources. 

Would it be possible to use the system class loader to load models instead of IOUtils class loader ?

I don't see any drawbacks of it but maybe I'm missing something.

Thank you.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,9 @@
 apply plugin: 'java'
 apply plugin: 'eclipse'
 apply plugin: 'application'
-
+apply plugin: 'maven'
+    
+group = 'com.github.ICIJ'
 // Gradle java plugin
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
--- a/src/edu/stanford/nlp/io/IOUtils.java
+++ b/src/edu/stanford/nlp/io/IOUtils.java
@@ -415,7 +415,7 @@ public class IOUtils  {
     // ms 10-04-2010:
     // - even though this may look like a regular file, it may be a path inside a jar in the CLASSPATH
     // - check for this first. This takes precedence over the file system.
-    InputStream is = IOUtils.class.getClassLoader().getResourceAsStream(name);
+    InputStream is = ClassLoader.getSystemResourceAsStream(name);
     // windows File.separator is \, but getting resources only works with /
     if (is == null) {
       is = IOUtils.class.getClassLoader().getResourceAsStream(name.replaceAll("\\\\", "/"));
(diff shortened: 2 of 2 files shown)


bamthomas:
oops sorry for the jipack commit

bamthomas:
If you want to dynamically load classes with a plugin system you have all the methods discussed [in this SO thread](https://stackoverflow.com/questions/60764/how-to-load-jar-files-dynamically-at-runtime).

To sum up you can :

* instantiate a new URLClassLoader, children of AppClassLoader (JDK>=9)  or main URLClassLoader (JDK<9) but is has drawbacks for example with singleton instances (see [the comments of this anwser](https://stackoverflow.com/a/60775/396155) )
* hack the visibility of the method addUrl of URLClassLoader with introspection (that doesn't work with JDK>=8 because it is the class AppClassLoader)
* pull a plugin system like OSGI, JCL class loader framework, XBean. I don't know how they are implemented. I haven't tested if it works with CoreNLP way of loading the models
* craft your own class loader that will allow to load jars and provide this implementation to the JVM property `java.system.class.loader`

So the benefit IMHO is that is gives more freedom to projects that integrates with CoreNLP to choose between these implementations, with model loading of CoreNLP that still works.

For Datashare (our software) it used to work with JDK8 because we implemented the 2nd solution. But as we wanted to be compatible with more recent JRE's we had to change the implementation (we used the 4th way), and then the models are not loading anymore.

So we patched CoreNLP and by now we are pointing to it as jitpack dependency, but it would be more convenient to use the official version.

AngledLuffa:
I tested this on Windows and Linux, and didn't see any problems, so I'll
merge it and hopefully it will continue to behave :)

bamthomas:
Thank you.

Just beware of the jitpack commit (with ICIJ group) : it you want that ppl can point to your master, you have to add the maven plugin to gradle and  the virtual group : 
```
apply plugin: 'maven'
group = 'com.github.stanford'
```
see https://jitpack.io/docs/BUILDING/

AngledLuffa:
I don't know why I didn't close this already, but

- I added a couple missing ClassLoader lines
- going to ignore the jitpack changes for now

AngledLuffa:
Hey, if you are able to do so, would you check out this change?

https://github.com/stanfordnlp/CoreNLP/commit/7be6ac36314b7fc3c972d870b0e58bfe3b37f50b

Someone found that the changed classloader was breaking their workflow, so I tried to include both.

PR #1133, opened 2021-02-07
replaced jaxb-core-2.3.0.1 with suitable istack implementation

grossDev:
you can only use one implementation of com.sun.xml.bind at a time;
fixed the naming of the jflex library

I make my contribution to the source files (.classpath, pom*.xml) under the same license as the Stanford CoreNLP project.
The added library "istack-commons-runtime" is created by SUN/Oracle and has the same dual licensing (GPL2 +CDDL) as the jaxb-api/impl libraries.


AngledLuffa:
Thanks!

PR #1140, opened 2021-03-07
Increase scripts/lexparser/lexparser.sh mem usage to 200m

mattmoore:
- While running the standalone parser 4.2.0 from https://stanfordnlp.github.io/CoreNLP/parser-standalone.html I noticed it crashing with test data included in `libexec/data/testsent.txt` with an out of memory exception. Increasing the value from 150m to 170m worked just fine for the test data. Set it to 200m in this patch for good measure.

- The 150m setting also causes test failures when trying to install via homebrew on mac, which runs the included `testsent.txt` file. I discovered this initially when creating a PR to bump the version to 4.2.0 for homebrew.

### DIFF
--- a/scripts/lexparser/lexparser.sh
+++ b/scripts/lexparser/lexparser.sh
@@ -10,5 +10,5 @@ fi
 
 scriptdir=`dirname $0`
 
-java -mx150m -cp "$scriptdir/*:" edu.stanford.nlp.parser.lexparser.LexicalizedParser \
+java -mx200m -cp "$scriptdir/*:" edu.stanford.nlp.parser.lexparser.LexicalizedParser \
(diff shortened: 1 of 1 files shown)


mattmoore:
While the PR description holds true, I was able to get around the homebrew test by using the `english-onesent.txt` sample, which unblocks my needs for now. Feel free to merge this or close it out if it is deemed unnecessary, as I can just easily edit the memory usage locally. My main concern for this PR was to allow homebrew installs to work.

AngledLuffa:
Mind changing it to dev instead of main?

mattmoore:
> Mind changing it to dev instead of main?

Sure thing!

PR #1182, opened 2021-09-05
Fixed Flaky Test

Auzel:
This tests assumes that there is a specific order that a HashMap and a Collection is iterated. However, the java documentation does not specify an order. So, if you run the _testAddRemove_ test in _CollectionValuedMapTest_ class a sufficient number of times, it will fail.

My fix is to sort the collection corresponding to each of the hash map keys. Then check if the map's toString() matches one of the permutations of its keys.

### DIFF
--- a/test/src/edu/stanford/nlp/util/CollectionValuedMapTest.java
+++ b/test/src/edu/stanford/nlp/util/CollectionValuedMapTest.java
@@ -5,6 +5,9 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


AngledLuffa:
What about just using the mechanism of the CollectionValuedMap by passing in a `MapFactory.ARRAY_MAP_FACTORY` and a `CollectionFactory.ARRAY_LIST_FACTORY`

Auzel:
I took your suggestion into account but I decided to check for equality between the two collections. It appears to be a simpler solution to the problem. Note that the new commit _9033a7b_ was done on my student account.

AngledLuffa:
That is really simple.  Love it

PR #1191, opened 2021-09-16
Italian

AngledLuffa:
Several minor improvements to make it easier to build models, along with scripts and a props file for Italian tokenizer, pos, ner, depparse models in CoreNLP


AngledLuffa:
Several minor improvements to make it easier to build models, along with scripts and a props file for Italian tokenizer, pos, ner, depparse models in CoreNLP

Issue #1184, opened 2021-09-07
Wrong ADP POS-Tag allocation in the german model 


Goldritter:
Hi,

I use both the stanford-corenlp and the german model in version "4.2.2" and found some odd behavior in the allocation of POS-Tags für words like "für" and "vor". These two words should be normally designed with "ADP" but I often get "NOUN" or even "PROPN" as POS-Tag for these words.

For example for the sentence "Welcher der Befunde ist für eine Gehirnerkrankung typisch?" "für" is tagged as "PROPN" an in "Welcher der Befunde ist am ehesten für hier am wahrscheinlichsten vorliegende Gehirnerkrankung typisch?" it is designated as "NOUN".

But for the sentences "Für wen ist das Essen?" and "Ich war reif für das Bett."  "für" is correctly tagged as "ADP". 

Here are the properties I used for the initialization:
(.setProperty properties "annotators" "tokenize, ssplit, mwt , pos, lemma, ner, parse, depparse")
(.setProperty properties "coref.algorithm", "neural")
(.setProperty properties "depparse.language", "german")
(.setProperty properties "ner.language", "de")
(.setProperty properties "tokenize.options" "untokenizable=noneDelete")

These are all the information I have. I hope they are helpfull.

Best regards
Goldritter

AngledLuffa:
I do not know any German, so I am unable to judge whether this is kinda of in the realm of possibility or just completely nuts.  In looking at the source data, there are a few examples where "für" is tagged as something other than ADP, which is doubtless why the model thinks it is a reasonable choice.  If these look wrong to you, we can hopefully fix the issue by fixing the data.

```
# sent_id = train-s8233
# text = Danach war er zwischen 1933 und 1956 als Chefarzt der Medizinischen Klinik am Kantonsspital Aarau unter anderem für_NOUN den Aufbau eines Grosslabors verantwortlich.
```

```
# sent_id = train-s8331
# text = Bei einem Bandwettbewerb im Film spielte McFly den Song auch mit seiner Band "Pinheads", der Vorsitzende der Jury, gespielt von Huey Lewis, nennt den eigenen Song daraufhin "eine \
Spur zu laut" für_NOUN den Schulwettbewerb.
```

```
# sent_id = train-s10753
# text = Typisch für_NOUN das Chuanqi ist eine dramatische Liebesgeschichte, die persönliche Schicksale und Machtkämpfe in der Politik darstellt.
```

```
# sent_id = train-s9560
[...]

PR #1202, opened 2021-10-09
Normalize all PTB produced tokens, not just the German ones

AngledLuffa:


AngledLuffa:
#1184

AngledLuffa:
This is causing a problem of turning NBSP into regular space.  Not sure that is actually a problem

PR #1241, opened 2022-01-12
Fix xml

AngledLuffa:


Haxatron:
line 198 should be spf.newSAXParser();

AngledLuffa:
oops... i edited out the part where the line started with SAXParser but ruined the whole thing!  thanks for catching that.

Issue #1195, opened 2021-09-29
How to get in touch regarding a security concern


zidingz:
Hey there!

I belong to an open source security research community, and a member (@srikanthprathi) has found an issue, but doesn’t know the best way to disclose it.

If not a hassle, might you kindly add a `SECURITY.md` file with an email, or another contact method? GitHub [recommends](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) this best practice to ensure security issues are responsibly disclosed, and it would serve as a simple instruction for security researchers in the future.

Thank you for your consideration, and I look forward to hearing from you!

(cc @huntr-helper)

AngledLuffa:
We'll take that under consideration - for now please just email <
***@***.***>

zidingz:
Hey @AngledLuffa

Thank you for your response. GitHub has a tendency to redact emails if supplied via email to a comments section. See how it appears on this issue.

<img width="455" alt="Screenshot 2021-09-30 at 16 25 18" src="https://user-images.githubusercontent.com/28839565/135484749-70aa3113-d1ab-4b11-8368-a8f0910b2c22.png">


If not a hassle, might you be able to try again? Else, simply insert it into this PR #1196. Thanks!

AngledLuffa:
Ha, that's funny.  I just posted the email address there.  Thanks for
preparing that

PR #1268, opened 2022-05-06
Update SECURITY.md

JamieSlome:
- Adding a qualified security policy

In relation to #1195


AngledLuffa:
I think contacting us directly will be sufficient

Issue #1291, opened 2022-08-04
NullPointerException with AnnotationSerializer (ProtobufAnnotationSerializer) and CoreMapExpressionExtractor


kschroeder:
I have a test that I run on `CoreMapExpressionExtractor` which relies on caching the result of an annotation run so I don't have to run the parser each time.  I use the `ProtobufAnnotationSerializer`.  However, when I unserialize the cached data all of the new line tokens have been set to `NULL`.  The result is that when I run the extractor a `NullPointerException` gets thrown at https://github.com/stanfordnlp/CoreNLP/blob/16ac6de8b1d5ecd959170ad78ea965ee5fba89a5/src/edu/stanford/nlp/ling/tokensregex/MatchedExpression.java#L321

Obviously, calling `containsKey` on a `null` causes the exception to be thrown.

The workaround that I used was simply to iterate over the tokens prior to using them and inserting a default new line `CoreLabel` element, and it seems to work.  Is there a better way to handle that?

AngledLuffa:
My new Star Trek theory is that all Andorians who go into engineering become Aenar:

- pale skin from years of being in computer labs and not getting any sun
- bad vision from reading technical manuals in the dark despite their mothers warning them not to
- psychic abilities to deal with support questions with no code samples

kschroeder:
I've _never_ read technical manuals in the dark.

This is the code for either pulling the data from cache or executing the annotator shown above.
```
    protected CoreDocument annotate(String filename) {
        try {
            CoreDocument fromCache = AnnotationCache.fromCache(filename); // <-- This is the next code block
            if (fromCache != null) {
                return fromCache;
            }

            File longerNameForFilename = new File(filename);
            String realpath = longerNameForFilename.getCanonicalPath();
            Annotation annotation = BasicAnnotator.basicAnnotation(ResourceLoader.loadFile(realpath)); // <-- This is the code block after that
            CoreDocument document = new CoreDocument(annotation);
            AnnotationSerializer serializer = new ProtobufAnnotationSerializer();
            serializer.writeCoreDocument(document, new FileOutputStream(AnnotationCache.getCacheFilename(filename)));
            return document;
        } catch (IOException | InvalidConfigurationException | ClassNotFoundException | NoSuchAlgorithmException e) {
     }
        return null; 
    }
```
Here is the code that actually the file (if it exists).. *This* is the code that returns the unserialized document that has the nulls in the token list.  It is referenced on the third line above.
```
[...]

PR #1295, opened 2022-08-12
Only run sentence splitting once, even if the user has manually added…

AngledLuffa:
Only run sentence splitting once, even if the user has manually added a redundant sentence splitter

Treat two character newlines as newlines


AngledLuffa:
https://github.com/stanfordnlp/CoreNLP/issues/1291

PR #1326, opened 2022-12-15
Server tregex

AngledLuffa:
Fix the server tregex endpoint so that the serialized trees method of input actually works.  Actually, the fix was lower down than the server - needed to update the serializer to attach the labels in the trees to the labels in the sentences


AngledLuffa:
(glanced at by @gangeli)

PR #1435, opened 2024-04-06
Update and rename examples/sample-maven-project/README.txt to Stanfor…

DharahasGoud:
…dCoreNLP


AngledLuffa:
Is this spam or something?  Why would we do this?

PR #1454, opened 2024-06-24
austrian german month names for german tokenizer

j3ernhard:
Austrian German month names for German tokenizer post processor:
https://en.wikipedia.org/wiki/Austrian_German


AngledLuffa:
Great, thanks for the contribution!

PR #1484, opened 2025-03-06
Semgrex contains

AngledLuffa:
Add functionality to semgrex to search for/against key/value pairs in fields on words which are maps.  Currently the one thing which this really applies to is morphofeatures.  When searching on morphofeatures, it matches a key string, then checks that the value does or doesn't match.

Current syntax:

```
{morphofeatures:{k1:/v1/;k2!:/v2/}}
```

represents a morphofeatures that must have k1=v1, but k2 != v2.  `/.../` represents regex, whereas not having `/.../` would force an exact match.


AngledLuffa:
Discussed syntax with @manning , happy with current status

PR #1485, opened 2025-03-17
added lemmas to tsv even without verbose output format option

churow:


src/edu/stanford/nlp/tagger/maxent/MaxentTagger.java
```diff
@@ -1435,8 +1435,16 @@ private static String getTsvWords(boolean verbose, boolean outputLemmas,
         throw new IllegalArgumentException("Expected HasTags, got " +
                                            hw.getClass());
       }
+      sb.append(word);
+      sb.append('\t');
+      if (outputLemmas) {
+        String lemma = ((HasLemma) hw).lemma();
```
AngledLuffa, line 1441: would you check that it is a `HasLemma` here first?

churow: I just followed original use of HasTag without check pattern.
HasWord, HasTag & HasLemma are implemented through src/edu/stanford/nlp/ling/AbstractToken.java

AngledLuffa: Yes, that makes sense.  However, the assumption is that every class that comes through the tagger will be both HasWord and HasTag, whereas if the tagger is using the WordTag class, there may not be a Lemma.

AngledLuffa:
I am drafting a new release this weekend, and aside from the change I requested I also need it merged to `dev` and not `main`, so I recreated the change and left you as the author:

https://github.com/stanfordnlp/CoreNLP/pull/1487

